PR #116, opened 2025-02-20
feat(logs): add `supportingQueryType` to identify requests done from plugin

svennergr:
Similar pattern used in logs drilldown.
https://github.com/grafana/logs-drilldown/blob/main/src/services/query.ts#L57


NWRichmond (approved):
Thanks for this!

PR #117, opened 2025-02-21
chore: exclude non-default provisioned plugins from source control

NWRichmond:
## What is this change?

This PR is a quality-of-life update. It allows us to provision datasources via a `provisioning/datasources/local.yaml` file (note: it doesn't need to be called `local`) without fear of committing any `secureJsonData[basicAuthPassword]`s to this repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,10 @@
 .env*
 !.env.example
 
+# Local provisioned datasources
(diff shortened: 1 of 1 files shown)


bohandley (approved):
Nice! LGTM! 🚀

PR #120, opened 2025-02-26
RelatedLogs: Display count of found logs in the Related Logs tab

### LINKED ISSUE #119: RelatedLogs: Provide logs count in the Related Logs tab
It would be helpful to have a count of Related Logs before users click on the Related Logs tab. For example:

Image

NWRichmond:
## Related issue

This PR closes #119.

## What does this do?

This PR extracts logic from the Related Logs scene to display a count of related logs in the Metric scene's action view tabs. Most of the orchestration of related logs is now handled in a new utility class (`src/RelatedLogs/RelatedLogsOrchestrator.ts`).

demo related logs count from breakdown

Other changes include:
- When the related logs feature is enabled, we now track the number of related logs that exist when a user clicks on an Action View tab. The related logs count is only reported when conditions have been met to show related logs (e.g. when at least one label filter has been applied, or the selected metric is the product of a Loki recording rule). 
- The `NoRelatedLogsScene` has a new look and updated messaging, which aims to improve clarity.
- The logs DS picker now filters out data sources with no log lines.
- The Grafana server created by `npm run server` wasn't previously able to connect to TSDBs created outside of our `docker-compose.yaml`. This PR updates the config to allow adding data sources with other TSDBs running on your `localhost`. It also now has the `exploreMetricsRelatedLogs` feature toggle enabled.

## Demo

https://github.com/user-attachments/assets/30ebe02d-e04e-4bba-b703-96adbd2d3827

## How to test

(Adapted from https://github.com/grafana/grafana/pull/98775)

I recommend using https://github.com/NWRichmond/demo-prometheus-loki-shared-labels to test this PR. If you'd like to test with other Prometheus & Loki resources, awesome!

1. Check out https://github.com/NWRichmond/demo-prometheus-loki-shared-labels and follow the Getting Started to spin up Prometheus and Loki resources on ports `9098` and `3108`, respectively
2. Create new Prometheus and Loki data sources that point to these services. The easiest way to do that is to create a `provisioning/datasources/labels-cross-reference.yaml` file containing:

```yaml
apiVersion: 1

datasources:
  - name: prometheus-labels-x-ref
    uid: prometheus-labels-x-ref
    type: prometheus
    access: proxy
    url: http://host.docker.internal:9098
    basicAuth: false
    withCredentials: false
    jsonData:
      manageAlerts: true
      prometheusType: Prometheus
      prometheusVersion: '3.2.1'
      cacheLevel: 'Low'
  - name: loki-labels-x-ref
    uid: loki-labels-x-ref
    type: loki
    access: proxy
    url: http://host.docker.internal:3108
    jsonData:
      manageAlerts: true
```

4. In Metrics Drilldown, select the new Prometheus data source (`prometheus-labels-x-ref`)
5. Select a metric and apply some labels to the filter. Some labels are shared (such as `region=us-west-2,environment=dev`) while others are not (`instance=demo-app:80`). We expect logs to only appear when the filters contain shared labels.

### DIFF
--- a/src/RelatedLogs/NoRelatedLogsFoundScene.tsx
+++ b/src/RelatedLogs/NoRelatedLogsFoundScene.tsx
@@ -27,8 +27,7 @@ export class NoRelatedLogsScene extends SceneObjectBase<SceneObjectState> {
           </ul>
         </Text>
         <Text variant="bodySmall" color="secondary">
-          Note: Related logs is an experimental feature that attempts to find logs with labels matching your current
-          metric selection.
+          Note: Related logs is an experimental feature.
         </Text>
       </Stack>
     );
--- a/src/RelatedLogs/NoRelatedLogsFoundScene.tsx
+++ b/src/RelatedLogs/NoRelatedLogsFoundScene.tsx
@@ -14,16 +14,16 @@ export class NoRelatedLogsScene extends SceneObjectBase<SceneObjectState> {
           We couldn&apos;t find any logs related to the current metric with your selected filters.
         </Alert>
         <Text>
-          To find related logs, you can try:
+          To find related logs, try the following:
           <ul className={styles.list}>
-            <li>Adjusting your label filters to include labels that exist in both the current metric and your logs</li>
+            <li>Adjust your label filters to include labels that exist in both the current metric and your logs</li>
             <li>
-              Selecting a metric created by a{' '}
+              Select a metric created by a{' '}
               <TextLink external href="https://grafana.com/docs/loki/latest/alert/#recording-rules">
                 Loki Recording Rule
               </TextLink>
             </li>
-            <li>Broadening the time range to include more data</li>
+            <li>Broaden the time range to include more data</li>
           </ul>
         </Text>
         <Text variant="bodySmall" color="secondary">
--- a/src/RelatedLogs/NoRelatedLogsFoundScene.tsx
+++ b/src/RelatedLogs/NoRelatedLogsFoundScene.tsx
@@ -11,19 +11,19 @@ export class NoRelatedLogsScene extends SceneObjectBase<SceneObjectState> {
     return (
       <Stack direction="column" gap={2}>
         <Alert title="No related logs found" severity="info">
-          We couldn't find any logs related to the current metric with your selected filters.
+          We couldn&apos;t find any logs related to the current metric with your selected filters.
         </Alert>
         <Text>
           To find related logs, you can try:
           <ul className={styles.list}>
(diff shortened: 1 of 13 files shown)


src/RelatedLogs/RelatedLogsOrchestrator.ts
```diff
@@ -0,0 +1,253 @@
...
+import { lokiRecordingRulesConnector } from '../Integrations/logs/lokiRecordingRules';
+import { type MetricScene } from '../MetricScene';
+
+type DataSource = DataSourceInstanceSettings<DataSourceJsonData>;
+
+/**
+ * Manager class that handles the orchestration of related logs functionality.
+ * This centralizes logs-related logic that was previously spread across multiple components.
+ */
+export class RelatedLogsOrchestrator {
```
bohandley, line 16: Love the name! Good refactoring here too

src/RelatedLogs/NoRelatedLogsFoundScene.tsx
```diff
@@ -1,28 +1,34 @@
...
-            <li>adjust the label filter to find logs with the same labels as the currently-selected metric</li>
+            <li>Adjusting your label filters to include labels that exist in both the current metric and your logs</li>
             <li>
-              select a metric created by a{' '}
+              Selecting a metric created by a{' '}
               <TextLink external href="https://grafana.com/docs/loki/latest/alert/#recording-rules">
                 Loki Recording Rule
               </TextLink>
             </li>
+            <li>Broadening the time range to include more data</li>
```
bohandley, line 26: Nice instructions! These are clear but I would go straight for the imperative:

To find related logs, try the following:
- Adjust your label filters...
- Select a metric...
- Broaden the time range...

[General conventions from the writer's toolkit here](https://grafana.com/docs/writers-toolkit/write/style-guide/style-conventions/)

bohandley (approved):
Looks great to me! Great job organizing the code, I was able to test the number of logs using the suggested data source as well as the otel-prometheus one (with otel-logs), and the "no logs" state looks good.

Just a small writer's toolkit nit, but otherwise 🚀 

Screenshot 2025-02-27 at 12 36 37 PM
Screenshot 2025-02-27 at 12 36 59 PM

src/RelatedLogs/NoRelatedLogsFoundScene.tsx
NWRichmond: Thank you for this! It's a nice upgrade.

Implemented in https://github.com/grafana/metrics-drilldown/pull/120/commits/a3a5f39a2b50c259c1965baf324adcdcf6cd6c3d ✅

bohandley (approved):

PR #165, opened 2025-03-11
feat(WingMan): Metrics tree filters prototype

grafakus:
Quick prototype for a very granular filters tree based on the metric name parts:

https://github.com/user-attachments/assets/e79e9004-d3ab-429b-9fe3-13d1ddb88f86

We could also use this component for bulk actions on a (large) list of metrics (e.g. configuring a Prometheus function / viz panel options).

**With the new sidebar:**

image

image


gtk-grafana:
Yes please! This is my biggest desire with Metrics Drilldown on my homelab box.
An idea to keep deeply nested values from overwhelming the user would be to add breadcrumbs of the parent nodes and remove all other context besides the first level children of the selected node by default, I'm dealing with a similar problem in Logs Drilldown:

https://github.com/user-attachments/assets/d3addf7d-9dee-41a1-a29c-b1691fafc002

grafakus:
Closing because stale. Created an issue to capture the idea: https://github.com/grafana/metrics-drilldown/issues/714

PR #166, opened 2025-03-11
chore: otel experience e2e tests

shelldandy:
Adds e2e test for turning on the otel experience and exploring around


grafakus (approved):
Nice one! Thank you Miguel! 🙌🏾

I like how you structured the tests with the `test.step` function. When running in local and in the reports, it gives a nice output. The only drawback I see is that we have to add some explanatory code whereas [using fixtures](https://playwright.dev/docs/test-fixtures#using-a-fixture) could lead to (among many benefits) a self-explanatory code. Let's talk!

bohandley (approved):
LGTM! Thank you for this work, @shelldandy!

PR #169, opened 2025-03-13
feat(*): Add plugin info menu in the page header

grafakus:
### ✨ Description

image

This PR also adds a basic guide for contributing to the repo. The missing piece is to be sure we set the right commit SHA in the new `src/version.ts` file.


NWRichmond (approved):
Beautiful! Thank you so much for this 😍

Issue #170, opened 2025-03-13
Metric name isn't interpolating correctly when navigating to Explore from panel extension point


NWRichmond:
## What's happening?

In the Metric scene, navigating to Explore via the "Open in explore" button is working fine, but when navigating to Explore via the panel extension point, the Explore query isn't interpolating the metric name:

https://github.com/user-attachments/assets/5ebb3ce4-51f0-45a9-a8ad-3bd7d8a7b938

## Workaround

To get to Explore from Grafana Metrics Drilldown, use the "Open in explore" button instead of the panel extension point.

PR #172, opened 2025-03-13
Bug fix: Fix explore link in panel using metricScene

bohandley:
Fixes https://github.com/grafana/metrics-drilldown/issues/170

The open in explore link that is in the panel (not the open in explore icon) was not interpolating the metric in the expression passed to the explore query editor. This fixes that.

How to Test:

1. Select a metric in GMD.
2. See the metric panel.
3. Find the hamburger dots in the large metric panel.
4. Click explore link from the dots.
5. See that the query open in explore query editor has the metric correctly interpolated.

Screenshot 2025-03-13 at 11 02 01 AM
Screenshot 2025-03-13 at 11 02 38 AM

### DIFF
--- a/src/Menu/PanelMenu.tsx
+++ b/src/Menu/PanelMenu.tsx
@@ -2,7 +2,6 @@ import { type DataFrame, type PanelMenuItem } from '@grafana/data';
 import { getPluginLinkExtensions } from '@grafana/runtime';
 import {
   getExploreURL,
-  SceneCSSGridItem,
   sceneGraph,
   SceneObjectBase,
   VizPanel,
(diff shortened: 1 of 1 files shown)


src/Menu/PanelMenu.tsx
```diff
@@ -49,8 +50,9 @@ export class PanelMenu extends SceneObjectBase<PanelMenuState> implements VizPan
           // removing legendFormat to get verbose legend in Explore
           delete query.legendFormat;
         });
-        const trail = getTrailFor(this);
-        exploreUrl = getExploreURL(panelData, trail, panelData.timeRange);
+        // use the metric scene to get the explore url with interpolated variables
+        const metricScene = sceneGraph.getAncestor(this, MetricScene);
```
NWRichmond, line 54: This works in the Metric scene, but unfortunately it doesn't work in the Metric Select scene, where we also display the explore link in the panel menu:

https://github.com/user-attachments/assets/a60c8f83-7cb0-4431-8e71-37d7d9a67b57

bohandley: Ok, I got it working for both metric select scene and metric scene!

NWRichmond (approved):
It's working beautifully in both contexts now. Thanks for this timely fix, @bohandley!

bohandley:
Thank you for the code pairing @NWRichmond!

PR #174, opened 2025-03-13
handle series endpoint bug with empty matcher

bohandley:
What happened? Sometimes the label group by options don't load and you see that the drop down only included the `none` option.

Why?
This bug occurs when a DS does not support the labels match[] parameter.

If a Prometheus DS does not support the labels match[] api, we must use the series endpoint instead. But, the series endpoint rejects empty matchers.

`"{__name__=~".+",$filters}"` is evaluated as empty.

So, we can use `getTagKeys` which is used in adhocVariables and will handle the empty matcher for the series endpoint. 

How to test:

1. Create a data source that does not support the labels match api.
2. You can do this by explicitly setting the data source to use the series endpoint
Screenshot 2025-03-13 at 5 10 35 PM

3. Next, start wingman using that data source.
4. See that the label group options load.

Screenshot 2025-03-13 at 5 13 57 PM

Future work for group by options: 
We will need to create a variable dependency where the label group by options are reset every time the data source changes. Currently, the options will not change if you change the DS.

### DIFF
--- a/src/WingmanDataTrail/Labels/LabelsDataSource.ts
+++ b/src/WingmanDataTrail/Labels/LabelsDataSource.ts
@@ -11,9 +11,12 @@ import {
 import { getDataSourceSrv } from '@grafana/runtime';
 import { RuntimeDataSource, sceneGraph, type DataSourceVariable, type SceneObject } from '@grafana/scenes';
 
-import { VAR_DATASOURCE } from 'shared';
+import { VAR_DATASOURCE, VAR_FILTERS } from 'shared';
+import { isAdHocFiltersVariable } from 'utils/utils.variables';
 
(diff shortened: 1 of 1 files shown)


src/WingmanDataTrail/Labels/LabelsDataSource.ts
NWRichmond, line 1: Looks good - any tips on how to test?

NWRichmond (approved):
Approving in the interest of reducing friction on the `wingman-ensemble` branch :)

src/WingmanDataTrail/Labels/LabelsDataSource.ts
bohandley: Updating test instructions in the description!

PR #180, opened 2025-03-17
Chore: Set Grafana version 11.6.0-82198 as oldest supported version

yangkb09:
Make our CI workflow only test against 11.6.0-82198 and newer

- [x] Update src/plugin.json’s minimum Grafana version dependency to 11.6.0-82198
- [x] Update the minimum version in the Requirements section of src/README.md to 11.6.0-82198

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -4,7 +4,7 @@ The Grafana Metrics Drilldown app provides a queryless experience for browsing P
 
 ## Requirements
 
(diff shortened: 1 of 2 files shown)


NWRichmond (approved):
Thanks so much, @yangkb09! As expected, this change allows our e2e tests to run only with our intended supported version:

demo e2e on 11 6 0-pre only

Issue #192, opened 2025-03-18
Units in panel


bohandley:
Show units in the panel

"I wonder if we can not only infer what aggregation operator to use on metrics, but also use the name of metrics to infer a unit. For example, if I filter for “bytes” using the dev data source, I get 2.1k hits!
"

"How cool would it be to add units on these? So instead of 4500000000 we see 4.5gb. This could be pretty cool and save some horizontal space inside the visualizations. I see a similar pattern occur for “seconds” as well."

https://raintank-corp.slack.com/archives/C03FZ9T50R1/p1741624548782329

PR #197, opened 2025-03-20
fix: display unit in metric panel y-axis

yangkb09:
## description
fixes https://github.com/grafana/metrics-drilldown/issues/192

the y-axis of the panels now displays units like 's' for seconds-based metrics such as CPU metrics, 'bytes' for byte-based metrics, etc. Grafana automatically formats these units appropriately (e.g., converting bytes to KB, MB, GB as needed).

changes in `units.ts`
- [x] add more unit types to `UNIT_MAP` (milliseconds, bits, percent, count)
- [x] expand `RATE_UNIT_MAP` to handle more rate metrics
- [x] refactor `getUnitFromMetric` to be more robust using several detection strategies
- [x] update `getUnit` and `getPerSecondRateUnit` to work with the metric name directly instead of requiring a pre-processed part

#### before changes

https://github.com/user-attachments/assets/9f9acb01-6802-470f-8e60-a9c5fc7bc3b7

#### after changes

https://github.com/user-attachments/assets/6ab62a09-bec0-4277-80af-2c31d1b32c17

### DIFF
--- a/src/autoQuery/units.ts
+++ b/src/autoQuery/units.ts
@@ -1,13 +1,27 @@
 export const DEFAULT_UNIT = 'short';
 export const DEFAULT_RATE_UNIT = 'cps'; // Count per second
 
-const UNIT_MAP: Record<string, string> = { bytes: 'bytes', seconds: 's' };
+// Enhanced unit mappings with more common unit types
+const UNIT_MAP: Record<string, string> = {
+  bytes: 'bytes',
+  seconds: 's',
+  milliseconds: 'ms',
+  bits: 'bits',
+  percent: 'percent',
+  count: 'short',
+};
+
 const UNIT_LIST = Object.keys(UNIT_MAP);
+
+// Enhanced rate unit mappings
 const RATE_UNIT_MAP: Record<string, string> = {
   bytes: 'Bps', // bytes per second
+  bits: 'bps', // bits per second
   // seconds per second is unitless
   // this may indicate a count of some resource that is active
   seconds: 'short',
+  count: 'cps', // counts per second
+  percent: 'percent', // percent stays percent even as a rate
 };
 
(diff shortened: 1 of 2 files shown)


src/autoQuery/units.ts
```diff
@@ -16,20 +28,74 @@ export function getUnitFromMetric(metric: string) {
...
+
+  // Special case for rate metrics
+  const metricLower = metricName.toLowerCase();
+  if ((metricLower.includes('rate') || metricLower.includes('per_second')) && metricPart === 'bytes') {
+    return 'Bps';
+  } else if ((metricLower.includes('rate') || metricLower.includes('per_second')) && metricPart === 'bits') {
+    return 'bps';
+  } else if (metricLower.includes('rate') || metricLower.includes('per_second')) {
+    return 'ops'; // Operations per second as a general fallback for rates
+  }
```
shelldandy, line 98: should those return strings be better as a constant somewhere so its easier to import on tests and what not

src/autoQuery/units.ts
```diff
@@ -16,20 +28,74 @@ export function getUnitFromMetric(metric: string) {
...
     }
   }
+
+  // Check for common patterns
+  if (metricLower.endsWith('_ms') || metricLower.includes('_ms_')) {
+    return 'milliseconds';
+  } else if (metricLower.endsWith('_count') || metricLower.includes('_count_')) {
+    return 'count';
+  } else if (metricLower.includes('percentage') || metricLower.includes('ratio')) {
+    return 'percent';
```
shelldandy, line 67: same here, i just get nervous when returning strings like that since it can be a bit more error prone in the future for tests or refactors

shelldandy (approved):
minor nits but lgtm

src/autoQuery/units.ts
yangkb09: ah good points! i replaced the strings with constants

PR #198, opened 2025-03-20
[WIP] OTel experience: refactor to get ready for WIngman

bohandley:
THIS IS A WIP

This branch will focus on cleaning up 

1. the initial check
2. local storage check
3. OTel resource calls

Possibly (stretch)
4. Make a POC runtime datasource
4. Refactor extra variables as part of the new runtime datasource


NWRichmond (commented):
I really appreciate that we're free of a `useEffect` in DataTrail! When I select a data source with OTel resource attributes, the OTel Experience continues to work as I'd expect.

There are some things I don't understand yet. I'm mildly troubled by `gdev-prometheus` being a special case - I'd like to understand this better before ✅

NWRichmond (approved):
I think it's okay to accept that special handling for `gdev-prometheus` helps us now, even though we don't yet understand why we need it. Let's figure it out when time allows, and in the meantime, embrace the changes that this PR brings.

Thanks for this, @bohandley!

bohandley:
@NWRichmond here is the issue for investigating this! https://github.com/grafana/metrics-drilldown/issues/244

PR #215, opened 2025-03-25
fix: specify branch on e2e-workflow

shelldandy:
...and we want to upload artifacts


NWRichmond (approved):
great catches, thanks!

bohandley (approved):
🚀

PR #219, opened 2025-03-27
chore: metrics e2e

shelldandy:
Based of: https://docs.google.com/document/d/1kQUkpgKv7ocDGbcrwhPWO-1fgI0bKygeubqE1P8-siA/edit?tab=t.0#heading=h.n352bd4bm30c

Adds fixtures and tests for:

🟩Metric
Selecting breakdown by label
🟩Breakdown
Selecting filters
🟩Actions
Open in explore
Copy url
Bookmark
Select new metric
🟩Native Histogram


tests/select-metric-view.spec.ts
```diff
@@ -47,4 +47,20 @@ test.describe('Select metric view', () => {
     await expect(metricSceneDetails.getByLabel('Copy url')).toBeVisible();
     await expect(metricSceneDetails.getByLabel('Bookmark')).toBeVisible();
   });
+
+  test('Filtering by Label', async ({ selectMetricView }) => {
+    await selectMetricView.setAdHocFilter('label with 📈', 'metrics');
+
+    await expect(selectMetricView.getByText('label with 📈 = metrics')).toBeVisible();
```
grafakus, line 54: Could we be more precise and create a method like `selectMetricView.assertAdHocFilters(...)`?

shelldandy: they are just content assertions like on the other tests on this file tho

grafakus: Indeed, IMO the advantage of creating a method are:
- Semantic & consistency: we have `setAdHocFilter`, it feels logical to have `assertAdHocFilters` 
- Precision: what if "label with 📈 = metrics" or appears elsewhere in the UI?

What advantage(s) do you see not to do so?

package.json
```diff
@@ -17,6 +17,7 @@
     "e2e:fast": "playwright test --retries=2 --workers=3 -x --trace on",
     "e2e:debug": "playwright test --debug -x --trace on /test",
     "e2e:watch": "playwright test --ui",
+    "e2e:codegen": "playwright codegen localhost:3000/drilldown",
```
grafakus, line 20:
```suggestion
    "e2e:codegen": "playwright codegen",
```

We can always pass parameters to the npm script so that it works for everyone in the team. For instance, to test WingMan I'm using:

`npm run e2ecodegen -- http://localhost:3001/a/grafana-metricsdrilldown-app`

tests/select-metric-view.spec.ts
```diff
@@ -43,8 +45,47 @@ test.describe('Select metric view', () => {
...
+  });
+
+  test('Bookmark', async ({ selectMetricView }) => {
+    const panelTitle = 'a.utf8.metric 🤘';
+    await selectMetricView.selectMetricPanel(panelTitle);
+    await selectMetricView.createBookmark();
+    await expect(selectMetricView.getByText('Bookmark created')).toBeVisible();
+    await selectMetricView.seeAllBookmarksFromAlert();
+    // One for recent exploration and one for bookmark
+    await expect(selectMetricView.getByRole('button', { name: panelTitle })).toHaveCount(2);
```
grafakus, line 83: What if there's a bug and 2 bookmarks are created instead of one? This test would pass. Why not be more precise and target the bookmarks container?

As we use fixtures, we could have a "hasBookmark(panelTitle)" method or similar, wdyt?

grafakus (changes requested):
Nice job! I've left a couple of comments that should be easy to address.

tests/select-metric-view.spec.ts
shelldandy: If 2 bookmarks are created instead of 1 then the count would be 3 

One recent trail and 2 bookmarks

e2e/fixtures/views/SelectMetricView.ts
```diff
@@ -51,6 +51,27 @@ export class SelectMetricView extends DrilldownView {
     return this.getControls().getByPlaceholder('Filter by label values');
   }
 
+  // TODO: This can probably be better by having getByGrafanaSelector
+  // https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/selecting-elements#select-field
+  async setAdHocFilter(label: string, value: string) {
```
grafakus, line 56:
```suggestion
  async setAdHocFilter(label: string, operator: string, value: string) {
```

Very simple change that we'll use when testing "Include/Exclude" filtering after grouping by label.

e2e/fixtures/views/SelectMetricView.ts
```diff
@@ -51,6 +51,27 @@ export class SelectMetricView extends DrilldownView {
...
+    await this.getAdHocFilters().click();
+    await this.page.keyboard.type(label);
+    await this.page.keyboard.press('Enter');
+    await this.page.keyboard.type('=');
+    await this.page.keyboard.press('Enter');
+    await this.page.keyboard.type(value);
+    await this.page.keyboard.press('Enter');
+  }
+
+  async assertAdHocFilter(label: string, panelTitles: string[]) {
```
grafakus, line 66: This does not seem to make sense: if we assert on the ad hoc filters, we need to focus only on this piece of UI:

image

So, for now, let's go with something like:

```js
async assertAdHocFilter(expectedFilters: string[]) {
    const appControls = this.getByTestId('app-controls');
  
    for (const expectedFilter of expectedFilters) {
      await expect(appControls.getByText(expectedFilter)).toBeVisible();
    }
}
```

(we can improve it later)

playwright.config.ts
```diff
@@ -48,13 +48,18 @@ export default defineConfig<PluginOptions>({
...
+        // Used by the Copy URL test
+        permissions: ['clipboard-read', 'clipboard-write'],
+      },
       dependencies: ['auth'],
     },
   ],
 });
 
-function getGrafanaUrl() {
+export function getGrafanaUrl() {
```
grafakus, line 62:
```suggestion
function getGrafanaUrl() {
```

I think it's only used in this file, right?

src/DataTrailCard.tsx
```diff
@@ -18,7 +18,7 @@ export type Props = {
 };
 
 // Helper function to truncate the value for a single key:value pair
-const truncateValue = (key: string, value: string, maxLength: number) => {
+export const truncateValue = (key: string, value: string, maxLength: number) => {
```
grafakus, line 21:
```suggestion
const truncateValue = (key: string, value: string, maxLength: number) => {
```

tests/select-metric-view.spec.ts
```diff
@@ -43,8 +45,47 @@ test.describe('Select metric view', () => {
...
-    await expect(metricSceneDetails.getByLabel('Bookmark')).toBeVisible();
+    await expect(metricSceneDetails.getByLabel(UI_TEXT.METRIC_SELECT_SCENE.OPEN_EXPLORE_LABEL)).toBeVisible();
+    await expect(metricSceneDetails.getByLabel(UI_TEXT.METRIC_SELECT_SCENE.COPY_URL_LABEL)).toBeVisible();
+    await expect(metricSceneDetails.getByLabel(UI_TEXT.METRIC_SELECT_SCENE.BOOKMARK_LABEL)).toBeVisible();
+  });
+
+  test('Filtering by Label', async ({ selectMetricView }) => {
+    await selectMetricView.setAdHocFilter('label with 📈', 'metrics');
+    await selectMetricView.assertAdHocFilter('label with 📈 = metrics', [
+      'a.utf8.metric 🤘',
```
grafakus, line 56: See my comment above about `assertAdHocFilter`: if we prevent methods to have too many responsibilities, we make the test easier to read, understand and maintain in the long term.

Let's add 2 specific assertions on the panels that should be present after setting the ad hoc filter.

playwright.config.ts
shelldandy: no, its used for the `Copy url` e2e test as well that way we have a consistent grafana url also for the tests

tests/select-metric-view.spec.ts
grafakus: What if there's a bug on no recent trails are created anymore?

The point I'm trying to make is that _we need to be specific about the outcome that the user sees after performing these actions_. They don't see 2 buttons, they see one bookmark being created.

Happy to chat more in our next 1:1 about this topic.

grafakus (dismissed):
Just a few more comments that should be very easy to address.

The most important is the one about the implementation of `assertAdHocFilter`, thank you!

playwright.config.ts
grafakus: You're right, sorry I missed it!

NWRichmond (approved):
These look like great quality-of-life changes! Thank you for this :)

e2e/fixtures/views/SelectMetricView.ts
```diff
@@ -131,9 +151,76 @@ export class SelectMetricView extends DrilldownView {
     return this.getByTestId(`data-testid Panel header ${panelTitle}`);
   }
 
+  async assertPanel(panelTitle: string) {
+    await expect(this.getPanelByTitle(panelTitle)).toBeVisible();
+  }
+
+  async assertPanelIsNativeHistogram(panelTitle: string) {
+    const panel = this.getPanelByTitle(panelTitle);
+    await expect(panel.getByTestId('header-container').getByText('Native Histogram')).toBeVisible();
```
bohandley, line 160: Love it! Will this be testable for the Metrics Reducer? CC @yangkb09

bohandley (approved):

bohandley (approved):
Fresh 🚀 Great work!

PR #220, opened 2025-03-27
chore(*): Upgrade Grafana to v11.6.0 + some maintenance

grafakus:


.github/pull_request_template.md
```diff
@@ -0,0 +1,14 @@
+### ✨ Description
```
grafakus, line 1: I believe that having a template helps us structure our PRs and helps our reviewers. I provided this template as a starting point and would love to hear your thoughts and suggestions for improvements.

docker-compose.yaml
```diff
@@ -1,10 +1,15 @@
...
   grafana:
+    container_name: 'grafana-gmd'
     extends:
       file: .config/docker-compose-base.yaml
       service: grafana
-    container_name: 'grafana'
+    build:
+      args:
+        grafana_version: ${GRAFANA_VERSION:-11.6.0}
+    ports:
```
grafakus, line 11: That was the only line in `docker-compose.override.yaml`, having less configuration files keeps things easy to understand.

docker-compose.yaml
```diff
@@ -1,10 +1,15 @@
...
 services:
   grafana:
+    container_name: 'grafana-gmd'
     extends:
       file: .config/docker-compose-base.yaml
       service: grafana
-    container_name: 'grafana'
+    build:
+      args:
+        grafana_version: ${GRAFANA_VERSION:-11.6.0}
```
grafakus, line 10: Without this line, the correct version would come from the `GRAFANA_VERSION` env variable (as defined in our our local `.env` file) or v11.5.1 for `.config/docker-compose-base.yaml`

package.json
```diff
@@ -20,8 +22,27 @@
...
-  "author": "Grafana",
-  "license": "Apache-2.0",
+  "packageManager": "npm@10.5.0",
+  "engines": {
+    "node": ">=22"
+  },
+  "dependencies": {
+    "@bsull/augurs": "^0.9.0",
+    "@emotion/css": "11.10.6",
+    "@grafana/data": "^11.6.0",
```
grafakus, line 32: The file has been reorganised a bit for scan-ability. The change: `@grafana/data`, `@grafana/runtime `, `@grafana/schema ` and `@grafana/ui ` point to `^11.6.0`

src/plugin.json
```diff
@@ -52,7 +52,7 @@
     }
   ],
   "dependencies": {
-    "grafanaDependency": ">=11.6.0-82198",
+    "grafanaDependency": ">=11.6.0",
```
grafakus, line 55: This fix this lint issue:
image

src/helpers/MetricDatasourceHelper.ts
```diff
@@ -15,7 +14,7 @@ import { getDataSourceSrv } from '@grafana/runtime';
 import { type DataTrail } from '../DataTrail';
 import { VAR_DATASOURCE_EXPR } from '../shared';
 
-function isPrometheusDatasource(ds: DataSourceApi<any>): ds is PrometheusDatasource {
```
grafakus, line 18: My knowledge of TS is limited so I couldn't fix the error differently:
```
src/helpers/MetricDatasourceHelper.ts:18:64 - error TS2677: A type predicate's type must be assignable to its parameter's type.
  Type 'PrometheusDatasource' is not assignable to type 'DataSourceApi<any, DataSourceJsonData, {}>'.
    Types of property 'importQueries' are incompatible.
```

.gitignore
```diff
@@ -55,6 +55,8 @@ ci/
 /playwright/.auth/
 
 # Editor
+.cursor
+.vscode
```
NWRichmond, line 59: I think it's helpful to keep `.vscode/settings.json` in source control, to promote consistent handling of things like `formatOnSave`. Perhaps we can do this:
 
```suggestion
.vscode
!.vscode/settings.json
```

docker-compose.override.yaml
NWRichmond, line 1: Removing this breaks support for running Grafana on port 3001. See https://github.com/grafana/metrics-drilldown/pull/98 and https://github.com/grafana/metrics-drilldown/pull/124 for context.

docker-compose.yaml
NWRichmond: See my comment on `docker-compose.override.yaml` for context on why we need this

package.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "grafana-metricsdrilldown-app",
   "version": "1.0.0-4",
+  "author": "Grafana",
+  "license": "Apache-2.0",
```
NWRichmond, line 5: To reflect our actual `LICENSE`:

```suggestion
  "license": "AGPL-3.0",
```

package.json
```diff
@@ -20,8 +22,27 @@
...
+  "packageManager": "npm@10.5.0",
+  "engines": {
+    "node": ">=22"
+  },
+  "dependencies": {
+    "@bsull/augurs": "^0.9.0",
+    "@emotion/css": "11.10.6",
+    "@grafana/data": "^11.6.0",
+    "@grafana/lezer-logql": "^0.2.7",
+    "@grafana/prometheus": "^11.6.0-223871",
```
NWRichmond, line 34: Might be a good time to upgrade to `@grafana/prometheus@latest` since it will include the changes we needed from the `11.6` prerelease.

src/helpers/MetricDatasourceHelper.ts
NWRichmond, line 1: I suggest undoing these changes and instead upgrading to `@grafana/prometheus@latest`. The error seems to be due to a mismatch between `@grafana/` package versions.

NWRichmond (changes requested):
It might be helpful to split some of these changes up into separate PRs. If that's a pain, we can keep them together, but I suggest some changes:

.gitignore
```diff
@@ -55,6 +55,8 @@ ci/
 /playwright/.auth/
 
 # Editor
+.cursor
```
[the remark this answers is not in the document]
NWRichmond: I'm drafting a PR today that adds a `.cursor/rules/<new-rule>.mdc`, and I think it'd be beneficial for everyone to have access to it because it adds domain knowledge. Should we perhaps add `!.cursor/rules` to accommodate this?

docker-compose.override.yaml
grafakus: It does not, it picks the value from the local `.env` file, this is what I have in mine:

`GRAFANA_PORT=3001`

And it works fine

.gitignore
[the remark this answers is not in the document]
grafakus: I guess it's a team decision, I'm using local rules that seem good but I didn't want to push them for all. Maybe we can still undo this change in your PR?

docker-compose.yaml
grafakus: Answered!

.gitignore
grafakus: Still, the changes made locally are going to be inside `settings.json` and I might commit them one day :(

In my experience, it's a personal decision to format on save. The most important thing the team can benefit from is is to have prettier and linting rules set up, and to make sure they are applied before opening a PR for review.

package.json
grafakus: Ok, this PR didn't change the version, it was already set to `^11.6.0-223871`. Out of curiosity, what are the changes we needed?

src/helpers/MetricDatasourceHelper.ts
grafakus: I will try, thank you!

.gitignore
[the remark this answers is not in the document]
NWRichmond: Sounds good! We could also include them in source control on a case-by-case basis.

.github/pull_request_template.md
NWRichmond: Looks like a great start

docker-compose.override.yaml
grafakus: Ok, sorry Nick!

I now understands that it worked, but both ports (3000 & 3001) were exposed. Thank you for sharing the related PRs (that I should have read before answering).

I added the `!override` tag and it exposes only one port now.

package-lock.json
NWRichmond, line 1: The `license` updated isn't here yet. Can you please commit the results of:

```
npm i --package-lock-only
```

NWRichmond (approved):
Thanks for those updates! I'm now able to run with `GRAFANA_PORT=3001` working as expected :)

Small chore before merging: let's update the lockfile.

package-lock.json
grafakus: Commit push, thank you!

PR #221, opened 2025-03-27
feat: Remove history component

bohandley:
### ✨ Description

This PR is to remove the history component.

No history                                                                                                                                               |  With History
:---------------------------------------------------------------------------------------------:|:-------------------------:
 | 

### 📚 Important info

Why are we removing the history? [Rational here](https://docs.google.com/document/d/1BU9Jnw3Zr3f1dqJu6UlaiQj5I10JoFVvsLTC9p_wbO4/edit?tab=t.0#heading=h.cz4xk8s4vu12)

Removing the component it easy.

Updating the logic is more complicated but not impossible. The logic is now changed to not store multiple steps in history, but instead only store the most up to date URL values.

This mostly impacts RECENT drilldowns so that they are stored in local storage as URL values.

1 regression: rudderstack for identifying a selected metric from the related metrics tab is now not being measured.

#### Recent
1. Navigate to GMD and select a metric or add a filter
2. Navigate away from GMD by selecting something in the Grafana menu like 'Explore'
3. This will update recent drill down explorations.
4. Return to the GMD home page.
5. Click the `recent` and have state be updated to the most recent action by the URL values

#### Bookmarks
1. Add a bookmark from the metric scene
2. Go to home page
3. Navigate from bookmark
6. Trail should update the state based on the associated book mark and have no history


shelldandy (approved):
-1,500 lines lets goo

grafakus:
That's beautiful 🤩 @bohandley ! 

Suggestion: what do you think about also reducing the page padding a little bit? From `16px 24px` to `8px 16px`:

| Before | After |
|  ---   |  ---  |
| image | image |

Just a suggestion, we can do it in a separate PR.

bohandley:
> That's beautiful 🤩 @bohandley !
> 
> Suggestion: what do you think about also reducing the page padding a little bit? From `16px 24px` to `8px 16px`:
> 
> Before	After
> image	image
> Just a suggestion, we can do it in a separate PR.

Yes! More real estate, more space for panels, more data!!!

bohandley:
Go @shelldandy! Next steps to get that browser history 💪

shelldandy:
@bohandley adding browser history has been added and merged to this branch!

e2e/config/playwright.config.local.ts
```diff
@@ -3,6 +3,7 @@ import { config } from './playwright.config.common';
 export default config({
   reporter: [
     ['list', { printSteps: true }],
-    ['html', { outputFolder: '../test-reports', open: 'on-failure' }],
+    ['html', { outputFolder: '../test-reports', open: 'never' }],
   ],
+  workers: 4,
```
shelldandy, line 8: image

grafakus (approved):
Removing code! 🥳🥳

The PR looks neat and I've manually tested several scenarios that all work. Nice job!

Issue #213, opened 2025-03-25
Extensions: G12 API removals


leventebalogh:
### Background
https://github.com/grafana/grafana/pull/102102 is removing the deprecated APIs related to extensions, and one of those removals is `getPluginLinkExtensions()`, which is used by this plugin (`grafana-lokiexplore-app`) [here](https://github.com/grafana/metrics-drilldown/blob/79cc5d5225d540b177a2e993ef20361637476557/src/Menu/PanelMenu.tsx#L122).

### Possible solution
Unfortunately in G12 we are planning to remove all the getter functions and are relying on the hooks only, which is unfortunate for Scenes, as it's mostly built using classes. **Do you have an idea how widely this extension (`"grafana-metricsdrilldown-app/investigation/v1"`) is used?** One possible solution could be if we would expose an observable. 🤔

PR #232, opened 2025-03-31
refactor(PanelMenu): Make extensions compatible with different Grafana versions

leventebalogh:
Fixes https://github.com/grafana/metrics-drilldown/issues/213.

### What changed?
Adds a condition to check which extensions APIs are available in the current grafana runtime.
[The new observable APIs are introduced in this PR.](https://github.com/grafana/grafana/pull/103063)

### Why necessary?
We are planning to phase out `getPluginLinkExtensions()` in G12.


src/Menu/PanelMenu.tsx
```diff
@@ -1,5 +1,6 @@
-import { type DataFrame, type PanelMenuItem } from '@grafana/data';
-import { getPluginLinkExtensions } from '@grafana/runtime';
+import { type DataFrame, type PanelMenuItem, type PluginExtensionLink } from '@grafana/data';
+// @ts-ignore Certain imports are not available in the dependant package, but can be if the plugin is running in a different Grafana version
+import { getObservablePluginLinks, getPluginLinkExtensions } from '@grafana/runtime';
```
svennergr, line 3:
```suggestion
// Certain imports are not available in the dependant package, but can be if the plugin is running in a different Grafana version.
// We need both imports to support Grafana v11 and v12.
// @ts-expect-error 
import { getObservablePluginLinks, getPluginLinkExtensions } from '@grafana/runtime';
```

svennergr (dismissed):

svennergr (approved):
LGTM, thank you Levi!

grafakus:
Tested with Grafana v11.6 and it still works but I see some 404 errors @svennergr:

image

@leventebalogh What's the simplest way to test with Grafana v12? I couldn't find any Docker image

grafakus:
Tested on Grafana v12:
image


It seems that the links can not be obtained because

```ts
 await lastValueFrom(getObservablePluginLinks({ extensionPointId, context }));
```

does not resolve. Any idea?

svennergr:
> Tested on Grafana v12: image
> 
> It seems that the links can not be obtained because
> 
> ```ts
>  await lastValueFrom(getObservablePluginLinks({ extensionPointId, context }));
> ```
> 
> does not resolve. Any idea?

I now realized, I only tested it with Grafana `main` and that was still using `getPluginLinkExtensions` 🙈 
Can confirm that the observable never "finishes" and thus `lastValueFrom` doesn't return.

leventebalogh:
Sorry @grafakus, it was my bad:
- the `lastValueFrom()` was timing out as the observable never gets completed (which actually makes sense). We should use `firstValueFrom()` or `.pipe(first()`. 🤦 
- also found a bug in the implementation in core, [here is a fix PR](https://github.com/grafana/grafana/pull/103367)

svennergr:
Was just about to suggest what @leventebalogh just committed or to `pipe(take(1))`. 
@leventebalogh I can take the change in logs.

grafakus:
Tested again and it works 🎉

grafakus (approved):

PR #248, opened 2025-04-07
feat(WingMan): Keep only side bar with filters

grafakus:
### ✨ Description

**Related issue(s):** `-`

As it's been decided to keep the version of WingMan with the side bar with filters, this PR removes the code related to the other variants, as well as the onboarding views. 

The only WingMan URL is: http://localhost:3001/a/grafana-metricsdrilldown-app/trail-filters-sidebar

### 📖 Summary of the changes

Removing code! See diff tab for specific comments.

### 🧪 How to test?

- The build should pass


NWRichmond (approved):
The `/trail-filters-sidebar` route continues to work as before, and the other Wingman routes are redirected as expected ✅ 

Thanks for this PR!

PR #257, opened 2025-04-07
feat(ErrorView): Provide a link for a clean page reload

grafakus:
### ✨ Description

**Related issue(s):** `-`

Whenever a runtime error occurs, we display a banner suggesting to the user to reload the page. But if the error comes from some incorrect values of the URL search parameters, reloading does not have any effect. 

This PR is an attempt to fix this situation by providing a link to reload the page properly:

| Before | After |
|  ---   |  ---  |
| image | image  |

In a future PR, we'll handle the specific case where a data source is set in the URL search parameters, but does not exist.

### 📖 Summary of the changes

The link generated keeps only the timerange parameters `from`, `to` and `timezone`.

### 🧪 How to test?

- Manually, provoke a runtime error and check that clicking on the "reload" link works

### DIFF
--- a/src/App/ErrorView.tsx
+++ b/src/App/ErrorView.tsx
@@ -1,19 +1,44 @@
 import { css } from '@emotion/css';
 import { type GrafanaTheme2 } from '@grafana/data';
-import { useStyles2 } from '@grafana/ui';
-import React from 'react';
+import { TextLink, useStyles2 } from '@grafana/ui';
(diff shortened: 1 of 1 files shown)


NWRichmond (approved):
Perfect ✨ 

https://github.com/user-attachments/assets/432d33ba-812f-4454-8d10-08e1b4dc7d1b

Thanks for this!

PR #171, opened 2025-03-13
test(EndToEnd): Proposal

grafakus:
Working doc: https://docs.google.com/document/d/1f7_VRmp2PXWLgYnLorQD6tRfeEH74_U588NnKfTrHvs/edit?tab=t.0

PR #261, opened 2025-04-08
test(EndToEnd): Support extra Playwright args

grafakus:
### ✨ Description

**Related issue(s):** follow-up of https://github.com/grafana/metrics-drilldown/pull/171

This PR provides a way to pass extra args to Playwright, which can be useful for (e.g.) regenerate screenshots or filtering the tests we want to execute.

More info on the CLI args: https://playwright.dev/docs/test-cli#reference

### 📖 Summary of the changes

The extra args are passed to the Docker container via the `PLAYWRIGHT_ARGS` environment variable.

### 🧪 How to test?

- Locally, try the command described in the documentation (see diff tab)


package.json
```diff
@@ -13,10 +13,6 @@
     "typecheck": "tsc --noEmit",
     "lint": "eslint --cache --ignore-path ./.gitignore --ext .js,.jsx,.ts,.tsx .",
     "lint:fix": "npm run lint -- --fix && prettier --write --list-different .",
-    "e2e": "playwright test",
```
grafakus, line 16: Unused

NWRichmond (commented):
After checking out the `grafakus/e2e-regen-screenshots` branch and running `npm run e2e:ci:server:up`, I ran the command in the docs:

```
PLAYWRIGHT_ARGS="select-metric-view.spec.ts -u" npm run e2e:ci
```

but after creating the containers, it exited with the following message:

```
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "select-metric-view.spec.ts": executable file not found in $PATH: unknown
```

NWRichmond (approved):
Thanks for those updates, @grafakus! The tests are running now.

The screenshot isn't being updated, but I believe it should be (the old screenshot contains three more rows of panels). For example:
difference between old and new metric select screenshot

Perhaps we need to tweak the `maxDiffPixelRatio`. I generated the diff above by setting the ratio to `0.0`. But we can look into this separately.

Thanks for this PR!

grafakus:
@NWRichmond Let's talk in our next call, "three more rows of panels" should be more than [1% diff pixel ratio](https://github.com/grafana/metrics-drilldown/blob/main/e2e/config/playwright.config.common.ts#L48) no? 🤔

NWRichmond:
> @NWRichmond Let's talk in our next call, "three more rows of panels" should be more than [1% diff pixel ratio](https://github.com/grafana/metrics-drilldown/blob/main/e2e/config/playwright.config.common.ts#L48) no? 🤔

Exactly. Looking forward to discussing more :)

PR #271, opened 2025-04-09
refactor(Variables): 

grafakus:
### ✨ Description

**Related issue(s):** follow-up of https://github.com/grafana/metrics-drilldown/pull/248

This PR is the refactor we discussed in our ensemble session.

### 📖 Summary of the changes

### 🧪 How to test?

All filters should behave as before:

1. **QuickSearch:**
  - When typing without any "Group by" option selected and with a "Group by" option selected
  - After typing: reloading the page should filter the metrics list
2. **Sidebar filters:** selecting different filters should work as expected
3. **Combining QuickSearch and the sidebar filters** should work as expected
4. **Combining QuickSearch, sidebar filters and "Group by"** should work as expected

In a future PR, we'll add some E2E tests to prevent any regression

### DIFF
--- a/src/DataTrail.tsx
+++ b/src/DataTrail.tsx
@@ -34,9 +34,6 @@ import React, { useEffect, useRef } from 'react';
 
 import { PluginInfo } from 'PluginInfo/PluginInfo';
 import { getOtelExperienceToggleState } from 'services/store';
-import { LabelsVariable } from 'WingmanDataTrail/Labels/LabelsVariable';
-import { FilteredMetricsVariable } from 'WingmanDataTrail/MetricsVariables/FilteredMetricsVariable';
-import { MetricsVariable } from 'WingmanDataTrail/MetricsVariables/MetricsVariable';
 
 import { NativeHistogramBanner } from './banners/NativeHistogramBanner';
 import { DataTrailSettings } from './DataTrailSettings';
@@ -785,9 +782,6 @@ function getVariableSet(initialDS?: string, metric?: string, initialFilters?: Ad
         placeholder: 'Select',
         isMulti: true,
       }),
-      new MetricsVariable({}),
-      new FilteredMetricsVariable(),
-      new LabelsVariable(),
     ],
   });
 }
--- a/src/WingmanDataTrail/GroupBy/MetricsWithLabelValue/MetricsWithLabelValueVariable.ts
+++ b/src/WingmanDataTrail/GroupBy/MetricsWithLabelValue/MetricsWithLabelValueVariable.ts
@@ -9,19 +9,16 @@ import {
   VAR_WINGMAN_SORT_BY,
   type SortingOption,
 } from 'WingmanDataTrail/HeaderControls/MetricsSorter/MetricsSorter';
-import { EventQuickSearchChanged } from 'WingmanDataTrail/HeaderControls/QuickSearch/EventQuickSearchChanged';
-import { QuickSearch } from 'WingmanDataTrail/HeaderControls/QuickSearch/QuickSearch';
-import { MetricsVariableFilterEngine } from 'WingmanDataTrail/MetricsVariables/MetricsVariableFilterEngine';
+import { EventMetricsVariableActivated } from 'WingmanDataTrail/MetricsVariables/EventMetricsVariableActivated';
+import { EventMetricsVariableDeactivated } from 'WingmanDataTrail/MetricsVariables/EventMetricsVariableDeactivated';
+import { EventMetricsVariableUpdated } from 'WingmanDataTrail/MetricsVariables/EventMetricsVariableUpdated';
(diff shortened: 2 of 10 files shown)


src/DataTrail.tsx
```diff
@@ -785,9 +782,6 @@ function getVariableSet(initialDS?: string, metric?: string, initialFilters?: Ad
         placeholder: 'Select',
         isMulti: true,
       }),
-      new MetricsVariable({}),
```
grafakus, line 788: We move them to the `MetricsReducer` where they are used (cohesion).

src/WingmanDataTrail/MetricsReducer.tsx
```diff
@@ -80,6 +96,65 @@ export class MetricsReducer extends SceneObjectBase<MetricsReducerState> {
     );
   }
 
+  private subscribeToFiltersEvents() {
```
grafakus, line 99: This is where the centralized filtering logic lives.

src/WingmanDataTrail/MetricsVariables/FilteredMetricsVariable.ts
```diff
@@ -8,95 +8,42 @@ import {
...
-        sideBar.subscribeToEvent(EventFiltersChanged, (event) => {
-          this.filterEngine.applyFilters({
-            [event.payload.type]: event.payload.filters,
-          });
-        })
-      );
-    } catch (error) {
-      console.warn('SideBar not found - no worries, gracefully degrading...', error);
-    }
+    this.publishEvent(new EventMetricsVariableActivated({ key: this.state.key as string }), true);
```
grafakus, line 37: What do you think about creating a helper to automatically decorate the class and publishes the 3 events? Something like:

```ts
addPublishFilterEvents(new FilteredMetricsVariable())
```

src/WingmanDataTrail/MetricsVariables/MetricsVariableFilterEngine.ts
```diff
@@ -122,11 +114,6 @@ export class MetricsVariableFilterEngine {
     return options.filter((option) => regexes.some((regex) => regex.test(option.value as string)));
   }
 
-  private notifyUpdate() {
-    // hack to force SceneByVariableRepeater to re-render
-    this.variable.publishEvent(new SceneVariableValueChangedEvent(this.variable), true);
```
grafakus, line 127: Not needed, no more hacks :)

src/WingmanDataTrail/GroupBy/MetricsWithLabelValue/MetricsWithLabelValueVariable.ts
```diff
@@ -34,6 +31,7 @@ export class MetricsWithLabelValueVariable extends QueryVariable {
     removeRules?: boolean;
   }) {
     super({
+      key: `${VAR_METRIC_WITH_LABEL_VALUE}-${labelName}-${labelValue}`,
```
NWRichmond, line 34: I believe we need to support UTF-8 label names here. Perhaps we can handle it upstream in `src/WingmanDataTrail/MetricsReducer.tsx` by wrapping the `labelName` passed to `MetricsGroupByList` with the `utf8Support` util from `@grafana/prometheus`:

```ts
(new MetricsGroupByList({ labelName: utf8Support(groupByValue) }) as unknown as SceneObjectBase),
```

src/WingmanDataTrail/MetricsReducer.tsx
NWRichmond, line 1: To avoid type assertions, I suggest using this pattern instead:

```ts
const labelName = variable.getValue();
if (typeof labelName === 'string') {
  // use the `labelName`
}
```

NWRichmond (commented):
I'll continue looking through this but wanted to share some initial feedback

src/WingmanDataTrail/MetricsVariables/FilteredMetricsVariable.ts
grafakus: Added in https://github.com/grafana/metrics-drilldown/pull/271/commits/adec5ee2fca26cd1c6b32c83a7e519247543dfa4

src/WingmanDataTrail/GroupBy/MetricsWithLabelValue/MetricsWithLabelValueVariable.ts
grafakus: AFAIK Scenes supports variable names with utf8 chars. Do you mean when we build the query in the `MetricsWithLabelValueVariable`?

NWRichmond: There's something broken in our handling of UTF-8 label names:

group by label value with utf8 label value error

but my suggestion doesn't fix it. Let's dig into this later, and ignore for now :)

NWRichmond (approved):
Wonderful refactor! Thank you for this. It's working nicely.

PR #276, opened 2025-04-10
refactor(Variables): Decouple metric variables from the UI component used for sorting

grafakus:
### ✨ Description

**Related issue(s):** follow-up of https://github.com/grafana/metrics-drilldown/pull/271

This PR is the 2nd part of the refactor we discussed in our ensemble session. 

### 📖 Summary of the changes

### 🧪 How to test?

Same as in https://github.com/grafana/metrics-drilldown/pull/271, but taking also in consideration sorting options

We'll add E2E tests in a future PR.

### DIFF
--- a/src/WingmanDataTrail/MetricsVariables/MetricsVariableFilterEngine.ts
+++ b/src/WingmanDataTrail/MetricsVariables/MetricsVariableFilterEngine.ts
@@ -1,4 +1,4 @@
-import { type QueryVariable, type VariableValueOption } from '@grafana/scenes';
+import { SceneVariableValueChangedEvent, type QueryVariable, type VariableValueOption } from '@grafana/scenes';
 import { cloneDeep, isEqual } from 'lodash';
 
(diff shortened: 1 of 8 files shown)


src/WingmanDataTrail/MetricsVariables/MetricsVariableFilterEngine.ts
```diff
@@ -121,4 +128,9 @@ export class MetricsVariableFilterEngine {
       return new RegExp('.*');
     }
   }
+
+  private notifyUpdate() {
+    // hack to force SceneByVariableRepeater to re-render
+    this.variable.publishEvent(new SceneVariableValueChangedEvent(this.variable), true);
```
grafakus, line 134: It appears we still need this hack. Without, the list of variables is not updated anymore when sorting/filtering.

NWRichmond (approved):
Beautiful refactor 😍

Sorting is working as expected, and there are no issues with changing the time range. Thank you so much, @grafakus!

PR #277, opened 2025-04-10
feat: bookmarks sidebar

shelldandy:
Fixes: https://github.com/grafana/metrics-drilldown/issues/253

image

Empty:

image


e2e/config/playwright.config.ci.ts
```diff
@@ -5,5 +5,5 @@ export default config({
   reporter: [['list'], ['html', { outputFolder: '../test-reports', open: 'never' }], ['github']],
   retries: 1,
   forbidOnly: true,
-  workers: 1,
+  workers: 2,
```
shelldandy, line 8: gotta go fast, but in CI :)

yangkb09:
miguel this is really good!! looks like bookmarks are being stored correctly in the sidebar. my main feedback is that it seems clicking on a bookmark doesn't take me to the metric

the rest of my feedback is non-blocking nits:
- the toast notif for successful bookmark still takes us to the old homepage (not sure if we want to do anything about this right now)
- would be nice to have the scrollbar be visible without needing to scroll
- the info icon on hover says "bookmarks". do we want to replace that with something else? or get rid of it
- the current logic for truncating bookmark names was tailored to the width of the `DataTrailCard`. @catherineymgui do we want to change how the truncation works given the width of the sidebar is smaller than the `DataTrailCard`s?


https://github.com/user-attachments/assets/e24bbc96-d505-4e54-b52e-53f174246c9f

https://github.com/user-attachments/assets/5aa6ac0c-9ee3-421f-bf39-c00eddbdb189

277:info

src/DataTrailCard.tsx
```diff
@@ -58,9 +60,11 @@ export function DataTrailCard(props: Props) {
   const { filters, metric, createdAt } = values;
 
   const heading = truncateValue('', getMetricName(metric), 39);
```
NWRichmond, line 62: per @yangkb09's commen about metric name truncation, this looks correct when set to `27`:

truncate set to 27

```suggestion
  const heading = truncateValue('', getMetricName(metric), 27);
```

src/WingmanDataTrail/SideBar/sections/BookmarksList.tsx
```diff
@@ -40,10 +43,40 @@ export class BookmarksList extends SceneObjectBase<BookmarksListState> {
...
     const styles = useStyles2(getStyles);
     const { title, description } = model.useState();
+    const { bookmarks } = getTrailStore();
+    const [_, setLastDelete] = useState(Date.now());
+
+    const onSelect = (index: number) => {
+      reportExploreMetrics('exploration_started', { cause: 'bookmark_clicked' });
+      const trail = getTrailStore().getTrailForBookmarkIndex(index);
+      getTrailStore().setRecentTrail(trail);
+    };
```
NWRichmond, line 53: Here's how we can navigate to the bookmarked metric:

Add imports:
```tsx
import React, { useContext, useState } from 'react';

import { MetricsContext } from 'App/App';
```

Use `goToUrlForTrail`:
```suggestion
    const [_, setLastDelete] = useState(Date.now());
    const { goToUrlForTrail } = useContext(MetricsContext);

    const onSelect = (index: number) => {
      reportExploreMetrics('exploration_started', { cause: 'bookmark_clicked' });
      const trail = getTrailStore().getTrailForBookmarkIndex(index);
      getTrailStore().setRecentTrail(trail);
      goToUrlForTrail(trail);
    };
```

With these changes, navigating to a bookmarked metric is working the way I'd expect:

https://github.com/user-attachments/assets/d5fea3c9-b2a0-494a-a539-e216ff8c46ae

NWRichmond (changes requested):
Looking good! The current lack of navigation to the bookmarked metric view is the only big blocker.

@yangkb09 left other helpful feedback; I'm echoing some of it here:

yangkb09:
clicking on a bookmark now takes me to the bookmarked metric - nice!! ❤️ 

i noticed a few other things. since i think the MVP functionality of bookmarks has been met (ability to bookmark, click on a bookmark and see the metric), i think these things can be addressed in the future. WDYT?

- a new filter icon is created on the top nav bar after i bookmark a metric

https://github.com/user-attachments/assets/b429f2e4-15ec-49b8-b2fe-2389cc40039b


- there's no "star" indicating a metric has already been bookmarked when i click into a bookmarked metric

https://github.com/user-attachments/assets/54e1d4da-3ef8-49aa-9346-eebf23f72bce


- bookmarks persist when i switch datasources (which may not be ideal since bookmarks are DS specific)

https://github.com/user-attachments/assets/24e774b1-abce-4cfd-84d7-077f2ee7d8ba

yangkb09 (dismissed):

catherineymgui:
This looks good! Two small nits:

1. Could we also use the star icon to represent bookmarks in the sidebar? So it matches the star action for bookmarking
2. Could we add more padding between the items in bookmarks? (Another 4px for example)

NWRichmond:
re:
> a new filter icon is created on the top nav bar after i bookmark a metric

This was my bad! Fixed by https://github.com/grafana/metrics-drilldown/pull/326.

NWRichmond (approved):

PR #278, opened 2025-04-10
fix(Otel): Handle error for a specific data source that doesn't handle errors li…

bohandley:
### ✨ Description

This is a hot fix for an otel call when calling a Prom version that errors when this happens:

1. resource call to `label_values` for deployment_environment
2. The label does not exist
3. Normally Prometheus versions do not error and will return no values
4. A specific vendor data source errors because it does not behave consistent with the others

[See this thread for testing.](https://raintank-corp.slack.com/archives/C0630AYC4SY/p1744299951706369?thread_ts=1744040899.244089&cid=C0630AYC4SY)


bossinc (approved):
Tested with Azure prom and Prom with azure creds. no more errors! it works great! thank you :tada:

NWRichmond (approved):
Thanks for this!

PR #282, opened 2025-04-11
MetricsReducer: Sorting enhancements with recents

### LINKED ISSUE #254: Sorting Enhancements
Sorting with support for recents (default), dashboards, alerts

NWRichmond:
### ✨ Description

**Related issue(s):**  Closes #254.

This PR makes some improvements to the sorting experience in the Metrics Reducer.

### 📖 Summary of the changes

- Adds a new **Default** sorting option, which sorts metrics alphabetically but puts the (up to) six most-recently selected metrics at the top of the list.
- Improves error handling for dashboard usage data fetching
- Allows the dashboard & alerting usage requests to encounter issues without resulting in an alert banner
- Removes the **Alphabetical (A-Z)** and **Reverse Alphabetical (Z-A)** sort options in favor o the new **Default** option
- Adds a description tooltip to the sorting options variable selector, to explain the available options

### 🧪 How to test?

After checking out the `254/sorting-enhancements-with-recents` branch and spinning up the project with `npm run server` and `npm run dev`, open the local Grafana instance and navigate to the `/a/grafana-metricsdrilldown-app/trail-filters-sidebar` route.

Next, try selecting a metric and returning to `/trail-filters-sidebar`. Notice that with the **Default** sorting, it appears at the top. This happens for the six most recently-selected metrics.

Note how when using filters, recently-selected metrics only appear when they sastisfy the conditions of the applied filters.

### Todo

I'm still working on e2e tests for the sorting behavior.


grafakus:
The feature works as expected, except for one the dashboards sorting. It looks like the query to get the `metricCounts` is aborted when landing:

image

src/WingmanDataTrail/HeaderControls/MetricsSorter/MetricsSorter.tsx
```diff
@@ -18,17 +18,76 @@ import { localeCompare } from 'WingmanDataTrail/helpers/localCompare';
...
+
+    // Remove the metric if it already exists and add it with new timestamp
+    const filteredMetrics = recentMetrics.filter((m) => m.name !== metricName);
+    filteredMetrics.unshift({ name: metricName, timestamp: now });
+
+    // Keep only the most recent metrics
+    const updatedMetrics = filteredMetrics.slice(0, MAX_RECENT_METRICS);
+    localStorage.setItem(RECENT_METRICS_STORAGE_KEY, JSON.stringify(updatedMetrics));
+  } catch (error) {
+    console.error('Failed to update recent metrics:', error);
```
grafakus, line 50: Can we use the [app logger](https://github.com/grafana/metrics-drilldown/blob/main/src/tracking/logger/logger.ts)?

So that we make sure they are reported to Faro.

src/WingmanDataTrail/HeaderControls/MetricsSorter/MetricsSorter.tsx
```diff
@@ -132,53 +197,75 @@ interface DashboardSearchItem {
...
+        limit: 500,
+      },
+      'grafana-metricsdrilldown-app-dashboard-search',
+      requestOptions
+    );
 
-      for (const panel of dashboard.panels) {
-        const { datasource } = panel;
-        if (!isPrometheusDataSource(datasource) || !('targets' in panel) || !panel.targets?.length) {
+    const metricCounts = await Promise.all(
```
grafakus, line 221: What if there are thousands of dashboards? To be cautious, can we use a library to limit the concurrency?
I believe https://www.npmjs.com/package/p-limit is a good fit.

src/WingmanDataTrail/HeaderControls/MetricsSorter/MetricsSorter.tsx
```diff
@@ -60,10 +121,14 @@ export class MetricsSorter extends SceneObjectBase<MetricsSorterState> {
 
   private activationHandler() {
     const sortByVar = sceneGraph.getVariables(this).getByName(VAR_WINGMAN_SORT_BY) as CustomVariable;
+    const deprecatedSortByOptions = new Set(['alphabetical', 'reverse-alphabetical']);
 
     this._subs.add(
       sortByVar.subscribeToState((newState, prevState) => {
-        if (newState.value !== prevState.value) {
+        if (deprecatedSortByOptions.has(newState.value as string)) {
```
grafakus, line 128: Nice to have thought about this! Can we generalise further and add some logic to set the value to `default` whenever the value from the URL search param is not known?

I've been trying to set the sort to (e.g.) `xxx` and it ended up in an inconsistent state:
image

NWRichmond: Sounds good! Tackled in https://github.com/grafana/metrics-drilldown/pull/282/commits/a6d5c49db72198344b75eba4f081b4f9dc2239a9.

src/WingmanDataTrail/HeaderControls/MetricsSorter/MetricsSorter.tsx
NWRichmond: yesss, let's! Fixed in https://github.com/grafana/metrics-drilldown/pull/282/commits/eeffb2a342c75c1faa339f12fa3bb28895ff6c16.

src/WingmanDataTrail/HeaderControls/MetricsSorter/MetricsSorter.tsx
NWRichmond: > What if there are thousands of dashboards?

We've set `limit: 500`, so there would be at most 500 requests for individual dashboards. In https://github.com/grafana/metrics-drilldown/pull/282/commits/0cbd28913d41d8e4297970e42316b7c7b1934a8a, I updated the fetching mechanism to deduplicate these requests (for example, when `DataTrail` re-renders excessively).

With that in mind, are you still concerned about the number of concurrent requests? If so, I'd rather go the route of offloading the requests to a worker than paginating the requests (which can get messy), but I'm open to feedback on this.

e2e/tests/metrics-reducer-view.spec.ts
```diff
@@ -11,4 +15,84 @@ test.describe('Metrics reducer view', () => {
     await metricsReducerView.assertSidebar();
     await metricsReducerView.assertMetricsList();
   });
+
+  test.describe('Metrics sorting', () => {
+    test('Default sorting shows recent metrics first, then alphabetical', async ({ metricsReducerView, page }) => {
```
grafakus, line 20: Have you considered screenshot testing?
Even if the UI changes in little bit in the next days, it would be easier to write these tests.

e2e/fixtures/views/MetricsReducerView.ts
```diff
@@ -140,4 +140,108 @@ export class MetricsReducerView extends DrilldownView {
     const panelsCount = await metricsList.locator('[data-viz-panel-key]').count();
     expect(panelsCount).toBeGreaterThan(0);
   }
+
+  async selectPrefixFilter(prefix: string) {
+    await this.assertSidebar();
+    const prefixFiltersSection = await this.page.getByTestId('metric-prefix-filters');
```
grafakus, line 146: We don't need to await for locators. Only if we perform an action on it or if we need to retrieve a value (e.g. its text content, ...)

src/WingmanDataTrail/MetricsVariables/MetricsVariableSortEngine.ts
```diff
@@ -16,11 +14,69 @@ export class MetricsVariableSortEngine {
   private variable: QueryVariable;
   private lastMetrics: string[];
   private sortBy?: SortingOption;
+  private internal: {
+    dashboardMetrics: Record<string, number>;
+    alertingMetrics: Record<string, number>;
+    dashboardMetricsPromise: Promise<Record<string, number>> | undefined;
```
grafakus, line 20: It feels we're mixing responsibilities: `MetricsVariableSortEngine` is responsible for sorting the variable options (domain logic), `fetchDashboardMetrics` is responsible for fetching the data (infrastructure).

If there is some logic related to in-flight requests (a promise already exists), why not adding it in `fetchDashboardMetrics` and `fetchAlertingMetrics`, to separate responsibilities and keep the code easier to read and maintain?

src/MetricSelect/WithUsageDataPreviewPanel.tsx
```diff
@@ -1,55 +1,71 @@
...
+  }
+
+  private _onActivate() {
+    const { metric } = this.state;
+    const metricsReducer = sceneGraph.getAncestor(this, MetricsReducer);
+    const filteredMetricsEngine = metricsReducer.state.enginesMap.get(VAR_FILTERED_METRICS_VARIABLE);
+    if (!filteredMetricsEngine) {
+      return;
+    }
+    filteredMetricsEngine.sortEngine.getDashboardUsageForMetric(metric).then((value) => {
```
grafakus, line 36: When reading this code, it does not feel right that the count comes from the sort engine, which should be responsible only for sorting (see my comment in `MetricsVariableSortEngine`).

src/WingmanDataTrail/ListControls/MetricsSorter/MetricsSorter.tsx
```diff
@@ -63,7 +130,10 @@ export class MetricsSorter extends SceneObjectBase<MetricsSorterState> {
 
     this._subs.add(
       sortByVar.subscribeToState((newState, prevState) => {
-        if (newState.value !== prevState.value) {
+        if (!this.supportedSortByOptions.has(newState.value as SortingOption)) {
```
grafakus, line 133: IMO we don't have to perform this logic every time the state changes.

I think it's enough to do it once, in the activation handler. The scenario being:
- the user lands on the page with a non-supported value in the URL search parameters
- the component mounts
- we detect the value is not supported => we change it to the default value
- after that, the only options that the user can choose come from the dropdown => we are safe

What do you tink?

src/WingmanDataTrail/ListControls/MetricsSorter/MetricsSorter.tsx
```diff
@@ -280,14 +403,36 @@ export function sortMetricsByCount(metrics: string[], counts: Record<string, num
...
 export function sortMetricsAlphabetically(metrics: string[]): string[] {
   return [...metrics].sort((a, b) => localeCompare(a, b));
 }
+
+/**
+ * Sort metrics with recent metrics first (by recency), then alphabetically
+ * @param metrics Array of metric names
+ * @returns Sorted array of metric names
+ */
+export function sortMetricsWithRecentFirst(metrics: string[]): string[] {
```
grafakus, line 419: This function could be a good candidate for unit testing in the future :)

src/WingmanDataTrail/HeaderControls/MetricsSorter/MetricsSorter.tsx
grafakus: 500 concurrent requests for each page load? It all depends on the server...

src/WingmanDataTrail/MetricsVariables/MetricsVariableSortEngine.ts
NWRichmond: I can consolidate the data fetching handling in the `fetch<Thing>Metrics` functions. The separation of concerns wasn't clear at first, as there was some data fetching in the `MetricVariableSortEngine` previously: https://github.com/grafana/metrics-drilldown/blob/24dc363dbed5d331893f8aa8c212de7949566c26/src/WingmanDataTrail/MetricsVariables/MetricsVariableSortEngine.ts#L68 https://github.com/grafana/metrics-drilldown/blob/24dc363dbed5d331893f8aa8c212de7949566c26/src/WingmanDataTrail/MetricsVariables/MetricsVariableSortEngine.ts#L79

src/WingmanDataTrail/MetricsReducer.tsx
```diff
@@ -98,6 +107,14 @@ export class MetricsReducer extends SceneObjectBase<MetricsReducerState> {
         this.state.drawer.close();
       })
     );
+
+    this._subs.add(
+      this.subscribeToEvent(MetricSelectedEvent, (event) => {
+        if (event.payload !== undefined) {
```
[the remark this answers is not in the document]
NWRichmond: There's a `new MetricSelectedEvent(undefined)` event in `MetricScene.tsx`. Although I don't expect that event to be caught by the Metrics Reducer, I think what we have here is fine.

e2e/tests/metrics-reducer-view.spec.ts
NWRichmond: Good ideas! Implemented ✅

src/WingmanDataTrail/MetricsVariables/MetricsVariableSortEngine.ts
NWRichmond: Done in https://github.com/grafana/metrics-drilldown/pull/282/commits/cc57fa356a60a1b462f6810737089c8cca0e42f7 ✅

src/WingmanDataTrail/ListControls/MetricsSorter/MetricsSorter.tsx
NWRichmond: Done ✅

src/WingmanDataTrail/ListControls/MetricsSorter/MetricsSorter.tsx
NWRichmond: Added in https://github.com/grafana/metrics-drilldown/pull/282/commits/ca15aca264ec7832a6c9c901b53e33677c9d2b5d ✅

src/WingmanDataTrail/HeaderControls/MetricsSorter/MetricsSorter.tsx
NWRichmond: I implemented the `p-limit` approach in https://github.com/grafana/metrics-drilldown/pull/282/commits/672c759249446189178457a29a28d4b14d49b593 ✅

NWRichmond:
@grafakus I've made many changes - I think it's a lot cleaner now!

PR #283, opened 2025-04-11
fix(Filtering): Fix filtering after clicking on the refresh button

grafakus:
### ✨ Description

**Related issue(s):** is caused by https://github.com/grafana/metrics-drilldown/pull/276

Before this PR, when a user would apply filters and then click on the refresh button, the filters would be ignored.

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- See https://github.com/grafana/metrics-drilldown/pull/276 for all the scenarios


src/WingmanDataTrail/MetricsReducer.tsx
```diff
@@ -147,26 +154,10 @@ export class MetricsReducer extends SceneObjectBase<MetricsReducerState> {
             prefixes: sideBar.state.selectedMetricPrefixes,
             suffixes: sideBar.state.selectedMetricSuffixes,
           },
-          { notify: false }
+          { forceUpdate: true, notify: false }
```
grafakus, line 157: The `forceUpdate` is the bug fix. After loading, we _always_ apply all the filters, then sort.

src/WingmanDataTrail/MetricsReducer.tsx
```diff
@@ -147,26 +154,10 @@ export class MetricsReducer extends SceneObjectBase<MetricsReducerState> {
...
-        const sortEngine = sortEnginesMap.get(event.payload.key)!;
-        const metricsSorter = sceneGraph.findByKeyAndType(this, 'metrics-sorter', MetricsSorter);
-        const sortByVariable = metricsSorter.state.$variables.getByName(VAR_WINGMAN_SORT_BY) as CustomVariable;
-
-        sortEngine.sort(sortByVariable.state.value as SortingOption);
-      })
-    );
-
-    this._subs.add(
-      this.subscribeToEvent(EventMetricsVariableUpdated, (event) => {
```
grafakus, line 162: This seemed to be a good idea but is not necessary in the end. We'll always sort, not whenever the variable options changes, but when we received a filter event of a sort event (see below). It's just... easier :)

src/WingmanDataTrail/MetricsVariables/helpers/areArraysEqual.ts
```diff
@@ -0,0 +1,4 @@
+import { isEqual } from 'lodash';
```
NWRichmond, line 1: Nice to see that `lodash` is one of the dependenvies that's externalized & [handled with SystemJS](https://grafana.com/developers/plugin-tools/key-concepts/npm-dependencies#dependency-sharing-mechanism), so we can use it freely without worrying about its impact on our bundle size :)

NWRichmond (approved):
Filtering + quick search is working nicely, before & after refreshing the page. Nicely done!!!

Issue #304, opened 2025-04-16
[RudderStack] Track Wingman usage


yangkb09:
Create new Rudderstack events for new Wingman features

Prioritize most important metrics/events so we can track high level engagement (don’t have to start out with a comprehensive implementation)

NWRichmond:
Closing this as complete, since the [subtasks](https://docs.google.com/document/d/12N97Oi8Vpkqr6puYYcErRKYsJN1fkk4_2UEalvWrvnU/edit?tab=t.0) are all ✅ ✅ ✅

PR #305, opened 2025-04-16
chore: create rudderstack event for sidebar toggle

yangkb09:
### ✨ Description

create rudderstack event for sidebar toggle

Ref https://github.com/grafana/metrics-drilldown/issues/304

### 📖 Summary of the changes

1. Added a new event type `metrics_sidebar_toggled ` to the `Interactions` interface in `src/interactions.ts` with these properties:
  - `action`: Either 'opened' or 'closed'
  - `section`: Optional string indicating which section was opened/closed
2. Updated the `SideBar` component to report the event in two places:
  - In the `setActiveSection` method when a section is opened or closed
  - In the close button's onClick handler to capture when the X button is clicked

### 🧪 How to test?

I added `console.log`s in the code locally and saw them appear when toggling the sidebar in the UI
Screenshot 2025-04-16 at 2 09 34 PM

https://github.com/user-attachments/assets/55188403-7211-4500-b7d0-88e121c21c13

### DIFF
--- a/src/WingmanDataTrail/SideBar/SideBar.tsx
+++ b/src/WingmanDataTrail/SideBar/SideBar.tsx
@@ -189,13 +189,7 @@ export class SideBar extends SceneObjectBase<SideBarState> {
               aria-label="Close"
               tooltip="Close"
               tooltipPlacement="top"
-              onClick={() => {
-                reportExploreMetrics('metrics_sidebar_toggled', {
-                  action: 'closed',
-                  section: visibleSection.state.key,
-                });
-                model.setActiveSection('');
-              }}
+              onClick={() => model.setActiveSection('')}
             />
             {/* TODO: find a better way */}
             {visibleSection instanceof MetricsFilterSection && <visibleSection.Component model={visibleSection} />}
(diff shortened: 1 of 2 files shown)


NWRichmond (commented):
Looks good! Perhaps we should avoid using the `wingman` name though...maybe we could use `metrics_sidebar_toggled` or something?

src/WingmanDataTrail/SideBar/SideBar.tsx
```diff
@@ -176,7 +189,13 @@ export class SideBar extends SceneObjectBase<SideBarState> {
...
               tooltip="Close"
               tooltipPlacement="top"
-              onClick={() => model.setActiveSection('')}
+              onClick={() => {
+                reportExploreMetrics('metrics_sidebar_toggled', {
+                  action: 'closed',
+                  section: visibleSection.state.key,
+                });
+                model.setActiveSection('');
+              }}
```
NWRichmond, line 198: I suggest undoing this to avoid double-reporting the close action. The action of closing is already caught in the `if (!sectionKey || sectionKey === visibleSection?.state.key) {` block above, where we're already reporting `action: 'closed'`.

```suggestion
              onClick={() => model.setActiveSection('')}
```

NWRichmond (changes requested):
One small tweak and I think we're good!

re: my suggestion, let me know if I'm missing something.

src/WingmanDataTrail/SideBar/SideBar.tsx
yangkb09: makes sense - thanks for catching that!

NWRichmond (approved):
Nice!

PR #315, opened 2025-04-21
fix(otel experience): OTel experience always defaults to off on drilldown start

bohandley:
### ✨ Description

This makes it so that OTel experience always defaults to off.

This prevents a bug when using `/trail-filters-sidebar` and a previous usage of OTel experience re-enables the feature, but there is now no more toggle switch.

### 📖 Summary of the changes

Removes the migration and always turns off the OTel experience from the start of

1. Url load
2. Recent or bookmark

### 🧪 How to test?

1. Use the original legacy flow for Metrics drilldown
2. Choose an otel friendly data source
3. Enable the otel experience
4. See that deployment environment is selected
5. Navigate to `http://localhost:3000/a/grafana-metricsdrilldown-app/trail-filters-sidebar` or whatever you use to check this code and use the Metrics Reducer
6. See that the deployment environment is not selcted as a filter.

https://github.com/user-attachments/assets/2263aea4-d6e2-44e5-a87e-3f10a80dfcf0


shelldandy (approved):
love this!

PR #319, opened 2025-04-22
fix: removing empty state start

shelldandy:
Fixes a bug where having a metric selected and then refreshing would take you to the reducer instead of the proper view


NWRichmond (approved):
🐛👋🏼

thanks for this!

bohandley (commented):
🎉🎉🎉

PR #326, opened 2025-04-23
fix: hide other metrics var by default

NWRichmond:
### ✨ Description

This fixes an issue where the "other metric filters" variable shows up when it's not supposed to.

### 📖 Summary of the changes

We hide the other metric filters variable by default.

### 🧪 How to test?

Have faith in this demo, which shows that the "other metric filters" variable doesn't show up as an empty combobox when navigating back to the Metrics Reducer from a selected metric:

https://github.com/user-attachments/assets/9b72c4c6-20c8-41c5-a32b-a73c7a853487


yangkb09 (approved):
🎉

PR #334, opened 2025-04-23
chore: deprecate bookmarks hp

shelldandy:
Helps with https://github.com/grafana/metrics-drilldown/issues/255

Before deprecating the homepage, lets deprecate bookmarks on the select metrics view

Removing from the normal experience:

image

image

Buuuut we keep it in the metrics reducer experience:

image

image


e2e/tests/select-metric-view.spec.ts
```diff
@@ -54,7 +54,7 @@ test.describe('Select metric view', () => {
     await expect(metricSceneDetails.getByLabel('Remove existing metric and choose a new metric')).toBeVisible();
     await expect(metricSceneDetails.getByLabel(UI_TEXT.METRIC_SELECT_SCENE.OPEN_EXPLORE_LABEL)).toBeVisible();
     await expect(metricSceneDetails.getByLabel(UI_TEXT.METRIC_SELECT_SCENE.COPY_URL_LABEL)).toBeVisible();
-    await expect(metricSceneDetails.getByLabel(UI_TEXT.METRIC_SELECT_SCENE.BOOKMARK_LABEL)).toBeVisible();
+    await expect(metricSceneDetails.getByLabel(UI_TEXT.METRIC_SELECT_SCENE.BOOKMARK_LABEL)).not.toBeVisible();
```
grafakus, line 57:
```suggestion
```

src/utils.ts
```diff
@@ -168,7 +168,7 @@ export function limitAdhocProviders(
   limitedFilterVariable.setState({
     getTagKeysProvider: async (
       variable: AdHocFiltersVariable,
-      currentKey: string | null
+      _: string | null // currentKey
```
grafakus, line 171: Why the change?

NWRichmond:
Let me know if I'm missing something, but I don't think we need to deprecate bookmarks from the homepage - we just need to deprecate the homepage itself. If we release the app in this state (pre-MetricsReducer, but with missing bookmarks), I think that'd be a confusing experience for users.

I believe the combination of https://github.com/grafana/metrics-drilldown/pull/337 and https://github.com/grafana/metrics-drilldown/pull/346 achieve our homepage deprecation goals, and I suggest removing any legacy bookmark code in https://github.com/grafana/metrics-drilldown/pull/346.

shelldandy:
Sounds good to me

PR #337, opened 2025-04-23
chore: homepage deprecation warning

shelldandy:
Supports: https://github.com/grafana/metrics-drilldown/issues/255

Uses localStorage to only show it once

image

### DIFF
--- a/src/banners/HomepageDeprecationBanner.tsx
+++ b/src/banners/HomepageDeprecationBanner.tsx
@@ -2,7 +2,7 @@ import { css } from '@emotion/css';
 import { Alert, useStyles2 } from '@grafana/ui';
 import React, { useEffect, useState } from 'react';
 
-const HP_BANNER_KEY = 'homepageDeprecationBanner';
+const HP_BANNER_KEY = 'metricsDrilldownHomepageDeprecationBanner';
 
 export const HomepageDeprecationBanner = () => {
   const [showBanner, setShowBanner] = useState(false);
--- a/src/banners/HomepageDeprecationBanner.tsx
+++ b/src/banners/HomepageDeprecationBanner.tsx
@@ -1,28 +1,39 @@
 import { Alert } from '@grafana/ui';
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 
 const HP_BANNER_KEY = 'homepageDeprecationBanner';
 
 export const HomepageDeprecationBanner = () => {
-  if (bannerHasBeenShown()) {
+  const [showBanner, setShowBanner] = useState(false);
+
+  useEffect(() => {
+    setShowBanner(!bannerHasBeenShown());
+  }, []);
+
+  if (!showBanner || bannerHasBeenShown()) {
     return null;
   }
 
   const onBannerRemove = () => {
-    bannerHasBeenShown();
+    setBannerHasBeenShown();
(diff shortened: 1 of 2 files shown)


catherineymgui:
I would agree with Zhehao's slack comment:
> Expand banner to the full width of the start your metrics exploration

Also:
> - Could we reword to include when the homepage will be removed? Example: "This homepage will be removed [as part of Grafana 12???], along with Recent Metrics. Bookmarks will move into the actions sidebar on the Metrics Select page."
> - Could we use the info variant (blue version)? Warning alerts are usually shown when there's an actual warning w/ the system

NWRichmond:
With respect to:
> as part of Grafana 12

How about "in Metrics Drilldown v1.0.0" instead? This might help users internalize that Metrics Drilldown isn't tied to Grafana's release cycle.

catherineymgui:
> How about "in Grafana Metrics Drilldown v1.0.0" instead? This might help users internalize that Metrics Drilldown isn't tied to Grafana's release cycle.

@zhehao-grafana what are your thoughts?

The only last thing is if we can have the text left-aligned instead of center-aligned? Thank you!

src/banners/HomepageDeprecationBanner.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <div className={styles.container}>
+      <Alert title="Homepage deprecation" severity="warning" elevated onRemove={onBannerRemove}>
+        <div className={styles.banner}>
+          This page will be removed soon.
+          <br />
+          Bookmarks will move alongside your list of metrics.
+        </div>
+      </Alert>
```
NWRichmond, line 32:
```suggestion
      <Alert title="Homepage deprecation" severity="info" elevated onRemove={onBannerRemove}>
        This page will be removed in Grafana Metrics Drilldown v1.0.0. Bookmarks will move alongside your list of
        metrics.
      </Alert>
```

src/banners/HomepageDeprecationBanner.tsx
```diff
@@ -0,0 +1,58 @@
...
+    container: css({
+      maxWidth: '904px',
+      width: '100%',
+      margin: '0 auto',
+      textAlign: 'center',
+    }),
+    banner: css({
+      maxWidth: '500px',
+      margin: '0 auto',
+    }),
```
NWRichmond, line 56:
```suggestion
    }),
```

NWRichmond (commented):
Capturing feedback:

demo deprecation banner with suggestions

src/banners/HomepageDeprecationBanner.tsx
```diff
@@ -0,0 +1,50 @@
+import { css } from '@emotion/css';
+import { Alert, useStyles2 } from '@grafana/ui';
+import React, { useEffect, useState } from 'react';
+
+const HP_BANNER_KEY = 'homepageDeprecationBanner';
```
NWRichmond, line 5: nit: could be helpful to namespace this

```suggestion
const HP_BANNER_KEY = 'metricsDrilldownHomepageDeprecationBanner';
```

NWRichmond (approved):
Looks good, and the storage aspect is working well! Thanks for this :)

NWRichmond (approved):

PR #339, opened 2025-04-24
Chore: Disable quick search clear button when there's no input

yangkb09:
### ✨ Description

Disable quick search clear button when there's no input

image

### 🧪 How to test?

Don't type anything into the search bar, see that the clear search button is disabled


NWRichmond (approved):
thank you for this!

PR #341, opened 2025-04-24
fix(MetricsReducer): Apply ad hoc filters when group by is active

grafakus:
### ✨ Description

**Related issue(s):** `.`

Before this PR, the ad hoc filters were not applied when a "group by" label had been selected.

| Before | After |
|  ---   |  ---  |
|image  | image |

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- Open the side bar to select a "group by" label
- Add some filters in the app header
- The metrics for each label should correspond to the selected filters

### DIFF
--- a/src/WingmanDataTrail/GroupBy/MetricsWithLabelValue/MetricsWithLabelValueDataSource.ts
+++ b/src/WingmanDataTrail/GroupBy/MetricsWithLabelValue/MetricsWithLabelValueDataSource.ts
@@ -56,7 +56,7 @@ export class MetricsWithLabelValueDataSource extends RuntimeDataSource {
     const timeRange = sceneGraph.getTimeRange(sceneObject).state.value;
     let metricsList: string[] = [];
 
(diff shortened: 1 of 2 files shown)


src/WingmanDataTrail/GroupBy/MetricsWithLabelValue/MetricsWithLabelValueDataSource.ts
```diff
@@ -56,7 +56,7 @@ export class MetricsWithLabelValueDataSource extends RuntimeDataSource {
     const timeRange = sceneGraph.getTimeRange(sceneObject).state.value;
     let metricsList: string[] = [];
 
-    let removeRules = query.startsWith('removeRules');
+    const removeRules = query.startsWith('removeRules');
```
grafakus, line 59: Not the most important ; I thought we had lint rules for using `const` over `let` in these situations.

NWRichmond (approved):
It's working well. Nice catch!

PR #346, opened 2025-04-24
chore: deprecate homepage

shelldandy:
Fixes: https://github.com/grafana/metrics-drilldown/issues/255

Deprecate Homepage - Default to wingman


grafakus (changes requested):
Shouldn't we leave the home route as it is and only replace the view that is rendered under `/a/%PLUGIN_ID%/`?

The `/trail-filters-sidebar` was a temporary pathname we used to test one of the 3 variants.

NWRichmond:
@shelldandy with https://github.com/grafana/metrics-drilldown/pull/378 merged, I think this PR can have a narrower scope, focusing on the removal of things that are no longer routable.

PR #349, opened 2025-04-24
chore: CVE fixes + update grafana deps to v12

shelldandy:
Fixing some medium CVEs that were starting to get low on SLO time

- simply ran `npm update`
- finally i had to add an alias to our webpack config to prevent this from happening (suggested by webpack itself)
image

image


src/WingmanDataTrail/GroupBy/MetricsWithLabelValue/MetricsWithLabelValueDataSource.ts
```diff
@@ -61,9 +61,9 @@ export class MetricsWithLabelValueDataSource extends RuntimeDataSource {
     const matcher = removeRules ? query.replace('removeRules', '') : query;
 
     if (MetricDatasourceHelper.datasourceUsesTimeRangeInLanguageProviderMethods(ds)) {
-      // @ts-expect-error: Ignoring type error due to breaking change in fetchSeriesValuesWithMatch signature
       metricsList = await ds.languageProvider.fetchSeriesValuesWithMatch(timeRange, '__name__', matcher);
     } else {
+      // @ts-expect-error: Ignoring type error due to breaking change in fetchSeriesValuesWithMatch signature
```
shelldandy, line 66: flipping the expectation since now we're using prometheus 12

package.json
```diff
@@ -27,81 +27,77 @@
...
+    "@grafana/ui": "^11.6.1",
     "@lezer/common": "^1.2.3",
     "p-limit": "^6.2.0",
-    "react": "18.2.0",
-    "react-dom": "18.2.0",
+    "react": "18.3.1",
+    "react-dom": "18.3.1",
     "react-router-dom": "^6.30.0",
-    "rxjs": "7.8.1",
+    "rxjs": "7.8.2",
```
NWRichmond, line 50: It's unfortunate that upgrading `rxjs` from `7.8.1` to `7.8.2` is leading to `as any` and `as unknown`s that weren't previously there. It appears this is due to `@grafana/data` version `11.6.1` having a dependency on `7.8.1`.

My suggestion is to first try upgrading the `@grafana/<thing>@11.6.1` packages to `@latest`, then see if we can continue to use the generics properly.

src/Menu/PanelMenu.tsx
```diff
@@ -138,7 +135,8 @@ const getInvestigationLink = async (addToExplorations: AddToExplorationButton) =
   // Check if we're running on Grafana v11
   if (config.buildInfo.version.startsWith('11.')) {
     try {
-      const { getPluginLinkExtensions } = await import('@grafana/runtime');
+      const runtime = await import('@grafana/runtime');
+      const getPluginLinkExtensions = (runtime as any).getPluginLinkExtensions;
```
shelldandy, line 139: New way of handling the dynamic import on Grafana 11

src/Menu/PanelMenu.tsx
```diff
@@ -1,7 +1,4 @@
 import { type DataFrame, type PanelMenuItem, type PluginExtensionLink } from '@grafana/data';
-// Certain imports are not available in the dependant package, but can be if the plugin is running in a different Grafana version.
-// We need both imports to support Grafana v11 and v12.
-// @ts-expect-error
```
shelldandy, line 4: Comment no longer valid on runtime v12

NWRichmond (approved):
Looks great! Thanks for this, @shelldandy 🤝

PR #353, opened 2025-04-26
MetricsReducer: Fix preview panel queries

### LINKED ISSUE #352: MetricsReducer: Preview panels don't use rate or groupings
Currently, the `MetricsVizPanel`'s use of `buildPrometheusQuery` is incomplete. It should correctly set `groupings` and `isRateQuery`, as we do in `VizPanel`.

NWRichmond:
### ✨ Description

**Related issue(s):** Fixes #352.

This PR fixes a couple of glaring issues with the query expressions used in `/trail-filters-sidebar` (as compared to those in `/trail`):
- groupings weren't being set for histograms
- queries that needed a [`rate()`](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate) didn't have a `rate()`

### 📖 Summary of the changes

The key change is that we are now correctly setting `groupings` and `isRateQuery` when calling `buildPrometheusQuery` in `src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx`. This leads to correct queries (compared to both the current `/trail`'s preview panels, and the preview panels from the legacy **Explore** > **Metrics** experience in Grafana 11.5).

Also, some refactoring:
- Fewer scene graph lookups for `DataTrail`
- Removed unnecessary items from state in `MetricVizPanel`

#### CI

- [x] All workflows should pass

#### Local

To test locally:

1. Check out the `352/fix-metric-viz-panel-queries` branch
2. Spin up the project with `npm run server` and `npm run dev`
3. Navigate to [`localhost:3001/a/grafana-metricsdrilldown-app/trail-filters-sidebar`](http://localhost:3001/a/grafana-metricsdrilldown-app/trail-filters-sidebar)
4. Observe that the panels (and their underlying queries) match the equivalent panels in `/trail`.


src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx
```diff
@@ -242,6 +234,23 @@ export class MetricVizPanel extends SceneObjectBase<MetricVizPanelState> {
...
+    const parts = metricName.split('_');
+    const suffix = parts.at(-1);
+
+    // Determine if this is a rate query based on metric suffix
+    const isRateQuery = rateQueryMetricSuffixes.has(suffix || '');
+
+    // Determine groupings based on metric suffix and native histogram status
+    let groupings: string[] | undefined;
+
+    if (isHistogram) {
```
bohandley, line 247: Have to be careful here, native histograms do not have the 'le' label and sometimes show no data when grouped by this

bohandley (changes requested):

src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx
bohandley: So it is working correctly for native histograms but in a strange way! I think this is ok for now but I want to add some context because it's a little long to follow.

1. Here we pass it the rate query when building the query runner
https://github.com/grafana/metrics-drilldown/blob/dc97df04f920230843d5fc721a4f42e994475154/src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx/#L158-L172

2. Then we identify it as a non rate query because it is missing the suffix.
https://github.com/grafana/metrics-drilldown/blob/dc97df04f920230843d5fc721a4f42e994475154/src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx/#L248

3. We still give it a grouping because it is a histogram
https://github.com/grafana/metrics-drilldown/blob/dc97df04f920230843d5fc721a4f42e994475154/src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx/#L253-L255

4. We begin building it as a non rate query with groupings but we pass it the original `prometheusFunction as NonRateQueryFunction` which is the `rate` function.
https://github.com/grafana/metrics-drilldown/blob/dc97df04f920230843d5fc721a4f42e994475154/src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx/#L218-L227

5. And finally we do not add `sum` and we do not add the `le` groupings in `getPromqlFunction` because `isRateQuery` is false
https://github.com/grafana/metrics-drilldown/blob/dc97df04f920230843d5fc721a4f42e994475154/src/autoQuery/buildPrometheusQuery.ts/#L77-L80

It works! 🤣 Native histograms are not `sum by (le)`

bohandley (approved):
Great catch for fixing the queries! I made a note for adding some code comments for native histograms, where they are identified as non rate queries, but still given the rate function as a `nonRateQueryFunction`. This is odd but shouldn't hold up the PR.

Thanks @NWRichmond for this work!

src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx
bohandley: Not requesting changes after looking deeper in to the code. 🚀

PR #361, opened 2025-04-29
fix(queries): add filters as matchers for each panel

bohandley:
### ✨ Description

Filters are not used in queries with matchers

Screenshot 2025-04-28 at 4 42 03 PM

### 📖 Summary of the changes

Get label filters from VAR_FILTERS to pass into the `MetricVizPanel` to use in queries

### 🧪 How to test?

See `duration_seconds` in the otel friendly data source. This metric requires a filter. In this example we see service_name.

Without this filter used ion the query, we get what looks like an empty query

Screenshot 2025-04-28 at 9 07 59 PM

With the filters in the query we get a normal looking heatmap

Screenshot 2025-04-28 at 9 08 19 PM

### DIFF
--- a/src/WingmanDataTrail/MetricsList/SimpleMetricsList.tsx
+++ b/src/WingmanDataTrail/MetricsList/SimpleMetricsList.tsx
@@ -16,7 +16,9 @@ import React from 'react';
 
 import { InlineBanner } from 'App/InlineBanner';
 import { WithUsageDataPreviewPanel } from 'MetricSelect/WithUsageDataPreviewPanel';
+import { VAR_FILTERS } from 'shared';
 import { getColorByIndex, getTrailFor } from 'utils';
+import { isAdHocFiltersVariable } from 'utils/utils.variables';
 import { LayoutSwitcher, LayoutType, type LayoutSwitcherState } from 'WingmanDataTrail/ListControls/LayoutSwitcher';
 import {
   VAR_FILTERED_METRICS_VARIABLE,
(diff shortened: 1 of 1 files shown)


src/WingmanDataTrail/MetricsList/SimpleMetricsList.tsx
```diff
@@ -27,6 +27,8 @@ import {
   MetricVizPanel,
 } from 'WingmanDataTrail/MetricVizPanel/MetricVizPanel';
 import { SceneByVariableRepeater } from 'WingmanDataTrail/SceneByVariableRepeater/SceneByVariableRepeater';
+import { VAR_FILTERS } from 'shared';
+import { isAdHocFiltersVariable } from 'utils/utils.variables';
```
NWRichmond, line 31: I'm getting a lint error with the order of these imports:
```
  30:1  error  `shared` import should occur before import of `utils`                                                        import/order
  31:1  error  `utils/utils.variables` import should occur before import of `WingmanDataTrail/ListControls/LayoutSwitcher`  import/order
```

It's fixed via:

```
npm run lint:fix
```

NWRichmond (approved):
I'm seeing labels in queries where expected now. Thanks so much!

src/WingmanDataTrail/MetricsList/SimpleMetricsList.tsx
NWRichmond: fixed in https://github.com/grafana/metrics-drilldown/pull/361/commits/b616ff9aa2a848eff5911c5f544a4cca67448eec

PR #364, opened 2025-04-29
fix(MetricScene): Hide prefix/suffix filters

grafakus:
### ✨ Description

**Related issue(s):** `-`

| Before | After |
|  ---   |  ---  |
|  https://github.com/user-attachments/assets/55606449-8dda-440a-b89c-63c0f3882174|  https://github.com/user-attachments/assets/76b1aaa4-7b35-4758-953a-0c1e55d5298e|

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- Go to the Metrics Reducer view
- Add some prefix/suffix filters
- The read-only filters input should appear next to the ad hoc filters
- Select a metric panel
- The read-only filters input should not appear

### DIFF
--- a/src/WingmanDataTrail/SideBar/SideBar.tsx
+++ b/src/WingmanDataTrail/SideBar/SideBar.tsx
@@ -110,7 +110,8 @@ export class SideBar extends SceneObjectBase<SideBarState> {
   }
 
   private onActivate() {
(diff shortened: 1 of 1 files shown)


NWRichmond (approved):
Thanks for adding the cleanup! 🧹

Issue #371, opened 2025-04-30
Fix zizmor issues


NWRichmond:
PRs that don't change GitHub actions are failing due to issues with GitHub actions, as reported by zizmor: https://github.com/grafana/metrics-drilldown/pull/368#issuecomment-2842360845

Let's fix these.

grafakus:
@NWRichmond Can we close this issue?

NWRichmond:
Good question - @shelldandy I see that https://github.com/grafana/metrics-drilldown/pull/372 was closed, but the latest zizmor reports have been fine...was this issue fixed by another PR?

shelldandy:
Oh yeah we can close this. 

This was fixed with 

https://github.com/grafana/metrics-drilldown/pull/386

Forgot to tag it in the PR description

PR #368, opened 2025-04-29
MetricsReducer: Isomorphic language provider methods

NWRichmond:
### ✨ Description

**Related issue(s):** Fixes #363.

This PR aims to:
- Fix an improper `ds.languageProvider.fetchLabelValues` call in `LabelsDataSource.fetchLabelValues`
- Make the "I need to support old and new flavors of `@grafana/prometheus`' `languageProvider` methods" approach simpler and more structured

### 📖 Summary of the changes

- Fix a bug in `src/WingmanDataTrail/Labels/LabelsDataSource.ts` in which the `ds.languageProvider.fetchLabelValues` method was called with arguments that didn't match either the old or new `fetchLabelValues` function signature
- Adds a new `MetricDatasourceHelper.datasourceUsesTimeRangeInLanguageProviderMethods` static method that documents the need for this check, why the solution isn't great, and avoids needing to implement the underlying `languageProvider` check elsewhere in application code
- Adds a simple e2e test that fails on `main` and passes on `363/isomorphic-language-provider-methods`
- Uses `@ts-expect-error` instead of `@ts-ignore`, so that when the error is fixed, TypeScript will be loud about telling us to remove the directive
- Some miscellaneous refactoring, including consolidating `isPrometheusDataSource` checks

#### Remote

- [x] All CI except zizmor should pass

#### Local

To confirm the fix of https://github.com/grafana/metrics-drilldown/issues/363 locally:

1. Check out the `363/isomorphic-language-provider-methods` branch
2. Spin up the Grafana 12 back end with `GRAFANA_IMAGE=grafana-dev GRAFANA_VERSION=12.0.0-240759 npm run server`
3. Spin up the front end with `npm run dev`
4. Navigate to [`localhost:3001/a/grafana-metricsdrilldown-app/trail-filters-sidebar`](http://localhost:3001/a/grafana-metricsdrilldown-app/trail-filters-sidebar)
5. Group by a label
6. You should see metrics panels grouped by labels, or "No metrics found for the current filters and time range." You shouldn't see a warning banner with the error shown in https://github.com/grafana/metrics-drilldown/issues/363.

PR #372, opened 2025-04-30
fix: zizmor permission issues

shelldandy:
Fixes: https://github.com/grafana/metrics-drilldown/issues/371

Fixes the `Playwright Published` action to make use of the plugin-ci-workflows action using dockerized pw

image


.github/workflows/playwright-published.yml
NWRichmond, line 1: While it'd be nice to use a single job here, my concern with using [`grafana/plugin-ci-workflows/.github/workflows/ci.yml`](https://github.com/grafana/plugin-ci-workflows/blob/main/.github/workflows/ci.yml) for this is that it'd create plugin artifacts in GCS every time a commit lands on `main`. That seems excessive, especially considering the size of the artifacts.

I wonder if we could instead take inspiration from  https://github.com/grafana/plugin-ci-workflows/blob/9bacf72016cc3cd1b0fdd0eacf10baccdaf4f87c/.github/workflows/ci.yml#L277-L285 and satisfy the [zizmor feedback](https://github.com/grafana/metrics-drilldown/pull/368#issuecomment-2842840373) about template expansion by leveraging the updated [`grafana/plugin-ci-workflows/actions/plugins/package@main`](https://github.com/grafana/plugin-ci-workflows/tree/main/actions/plugins/package).

If I'm thinking about this correctly, this approach would sort out the security piece without creating plugin artifact bloat.

.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ permissions:
 
 jobs:
   CI:
-    uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@main
+    uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@9bacf72016cc3cd1b0fdd0eacf10baccdaf4f87c
```
NWRichmond, line 21: We can pin the version of this workflow, but I don't think we need to for zizmor reasons. When we run https://github.com/grafana/github-action-patcher on this repo, it doesn't complain about this workflow being `@main`.

shelldandy: roger that ill revert to main

shelldandy:
Waiting on: https://github.com/grafana/deployment_tools/pull/257353

PR #373, opened 2025-04-30
chore: track sort by usage in the metrics reducer

NWRichmond:
### ✨ Description

**Related issue(s):** This PR supports #304.

The goal of this PR is to extend event tracking for `sortBy`, to capture sorting changes in the Metrics Reducer. Previously, we were tracking `sortBy` in the context of the breakdown scene, when selecting a label name (not "All"):

demo sort by in breakdown

### 📖 Summary of the changes

This PR adapts the `'sorting_changed'` interaction to deal with sorting changes in both the label breakdown scene and the metrics reducer. By using a [discriminated union](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#discriminated-unions) in `src/interactions.ts`, we can support both Sort By controls (each with their own options) with great type safety (`sortBy` values match the options in each scene).

### DIFF
--- a/src/Breakdown/ByFrameRepeater.tsx
+++ b/src/Breakdown/ByFrameRepeater.tsx
@@ -16,6 +16,7 @@ import { Alert, Button } from '@grafana/ui';
 import React from 'react';
 
 import { BreakdownSearchReset } from './BreakdownSearchScene';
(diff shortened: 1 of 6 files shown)


grafakus (approved):
More tracking to understand better our users! 🥳

PR #375, opened 2025-05-01
chore: create rudderstack event for quick search

yangkb09:
### ✨ Description

create rudderstack event for quick search

event only fires after a user has stopped typing for 1 second to prevent firing the event multiple times for a single search

ref https://github.com/grafana/metrics-drilldown/issues/304

### 📖 Summary of the changes

1. added new event type `quick_search_used` to `Interactions` interface
2. report the event when user types into quick search bar 

### 🧪 How to test?

i added `console.logs` in the code locally and saw them appear when typing into the quick search bar. you can see that typing quickly (alert) and slowly (kube) fires the event one time each
image

https://github.com/user-attachments/assets/e753e42c-3eaa-4dc9-874f-8964e00a7282

### DIFF
--- a/src/interactions.ts
+++ b/src/interactions.ts
@@ -112,8 +112,7 @@ export type Interactions = {
     )
   };
   // User types in the quick search bar
-  quick_search_used: {
-  };
+  quick_search_used: {};
   sorting_changed: {
     // By clicking on the sort by variable in the metrics reducer
     from: 'metrics-reducer',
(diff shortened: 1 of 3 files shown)


src/interactions.ts
```diff
@@ -111,6 +111,9 @@ export type Interactions = {
       | 'close'
     )
   };
+  // User types in the quick search bar
+  quick_search_used: {
+  };
```
NWRichmond, line 116:
```suggestion
  quick_search_used: {};
```

NWRichmond (dismissed):
Nice!

PR #384, opened 2025-05-02
refactor: generic health checks

NWRichmond:
### ✨ Description

This PR refactors a health check function (currently used for the Related Logs feature) in the following ways:
- It's more generic, and not tied to Loki-type data sources
- Switches to the API endpoint from the [deprecated `id`-based endpoint](https://grafana.com/docs/grafana/latest/developers/http_api/data_source/#get-a-single-data-source-by-id) to its [`uid`-based counterpart](https://grafana.com/docs/grafana/latest/developers/http_api/data_source/#get-a-single-data-source-by-uid)
- Adds caching to prevent superfluous data source fetching & health checks

### DIFF
--- a/src/DataTrail.tsx
+++ b/src/DataTrail.tsx
@@ -65,6 +65,7 @@ import {
 } from './shared';
 import { getTrailStore } from './TrailStore/TrailStore';
 import { currentPathIncludes, getTrailFor, limitAdhocProviders } from './utils';
+import { DataSourceFetcher } from './utils/utils.datasource';
 import { isSceneQueryRunner } from './utils/utils.queries';
 import { getSelectedScopes } from './utils/utils.scopes';
 import { isAdHocFiltersVariable, isConstantVariable } from './utils/utils.variables';
@@ -105,6 +106,8 @@ export interface DataTrailState extends SceneObjectState {
   nativeHistogramMetric: string;
 
   trailActivated: boolean; // this indicates that the trail has been updated by metric or filter selected
+
+  datasourceFetcher: DataSourceFetcher;
 }
 
 export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneObjectWithUrlSync {
@@ -136,6 +139,7 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
       histogramsLoaded: state.histogramsLoaded ?? false,
       nativeHistogramMetric: state.nativeHistogramMetric ?? '',
       trailActivated: state.trailActivated ?? false,
+      datasourceFetcher: state.datasourceFetcher ?? new DataSourceFetcher(),
       ...state,
     });
 
@@ -146,7 +150,7 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
     this.setState({ trailActivated: true });
 
     if (!this.state.topScene) {
-      this.setState({ topScene: getTopSceneFor(this.state.metric) });
+      this.setState({ topScene: this.getTopSceneFor(this.state.metric) });
     }
 
     // Some scene elements publish this
@@ -352,7 +356,7 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
     // must pass this native histogram prometheus knowledge deep into
     // the topscene set on the trail > MetricScene > getAutoQueriesForMetric() > createHistogramMetricQueryDefs();
     stateUpdate.nativeHistogramMetric = nativeHistogramMetric ? '1' : '';
-    stateUpdate.topScene = getTopSceneFor(metric, nativeHistogramMetric);
+    stateUpdate.topScene = this.getTopSceneFor(metric, nativeHistogramMetric);
 
     return stateUpdate;
   }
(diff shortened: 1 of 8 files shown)


src/utils/utils.datasource.ts
```diff
@@ -13,3 +17,37 @@ export function isPrometheusDataSource(input: unknown): input is PrometheusDatas
...
+        })
+        .then((health) => (health?.status === 'OK' ? healthyDataSources.push(ds) : unhealthyDataSources.push(ds)))
+        .catch(() => unhealthyDataSources.push(ds))
+    )
+  );
+
+  if (verbose && unhealthyDataSources.length) {
+    // Why not use `logger.warn` here? While this information might be useful for observant users
+    // who open DevTools, it's not an actionable insight for Grafana Metrics Drilldown developers.
+    console.warn(
```
grafakus, line 45: Can we use [the logger](https://github.com/grafana/metrics-drilldown/blob/main/src/tracking/logger/logger.ts#L66-L72) instead'

src/utils/utils.datasource.ts
```diff
@@ -13,3 +17,37 @@ export function isPrometheusDataSource(input: unknown): input is PrometheusDatas
     typeof input.uid === 'string'
   );
 }
+
+export async function findHealthyDataSources(type: 'prometheus' | 'loki', verbose = false) {
```
grafakus, line 21: How many times can this function called during a typical user session? My concern is that we call it more times than needed. And also, on larger instances, the `Promise.all` L30 could impact negatively the performance of the app.

If you find also my concerns valid, can we cache the results of this function?

src/utils/utils.datasource.ts
NWRichmond: My concern is that it's just noise in Frontend O11y. I'm not sure how it'd be valuable to us, but let me know if you see it differently.

src/utils/utils.datasource.ts
NWRichmond: For sure! Caching implemented in https://github.com/grafana/metrics-drilldown/pull/384/commits/d62c2148c769854877194f0c60f49713b946aac7 ✅

src/utils/utils.datasource.ts
grafakus: I'd argue that an increasing number of unhealthy DS is a valuable insight for GME developers. I don't think it'll create a lot of noise - my assumption being that unhealthy DS is not common.

From the code design perspective, the idea of having a single logger is to centralize the management of all log messages. It feels we're going backwards by using the console.

Also, if the info logged is neither actionable nor useful for GMD developers, why not displaying some information in the UI for the user and removing the `console.warn`?

NWRichmond: I'm open to `logger.warn`ing this for now, but I remain doubtful that this signal provides value to anyone but the user or the admins of the user's Grafana instance.

> why not displaying some information in the UI for the user and removing the console.warn?

I'd prefer this route, provided that we make it not too annoying. For example, I believe it'd be too annoying to display an alert banner every time unhealthy data sources are detected. A tooltip or something subtle next to the data source selector could be great. Unless the user happens to be an Admin and also happens to have control over the affected data source, it's usually not something that they should need to think about.

NWRichmond: Because I'd like to keep UI changes out of the scope of this PR, I shifted to using `logger` in https://github.com/grafana/metrics-drilldown/pull/384/commits/e919e98e2427607841d9dfa5f4144975d5ec6275

src/utils/utils.datasource.ts
```diff
@@ -13,3 +18,89 @@ export function isPrometheusDataSource(input: unknown): input is PrometheusDatas
...
+          if (health?.status === 'OK') {
+            healthyDataSources.push(ds);
+          } else {
+            unhealthyDataSources.push(ds);
+          }
+          // NOTE: This appears to be a false positive from sonarjs
+          // eslint-disable-next-line sonarjs/no-ignored-exceptions
+        } catch (error) {
+          unhealthyDataSources.push(ds);
+        }
```
grafakus, line 92: Actually SonarJS is right. Have you tried:

```suggestion
        } catch {
          unhealthyDataSources.push(ds);
        }
```
?

src/DataTrail.tsx
```diff
@@ -136,6 +139,7 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
       histogramsLoaded: state.histogramsLoaded ?? false,
       nativeHistogramMetric: state.nativeHistogramMetric ?? '',
       trailActivated: state.trailActivated ?? false,
+      datasourceFetcher: state.datasourceFetcher ?? new DataSourceFetcher(),
```
grafakus, line 142: Why do we instantiate `DataSourceFetcher` here and pass it around? Could we just create a singleton in `utils/datasource.ts` and import it where it's needed?

src/utils/utils.datasource.ts
NWRichmond: it works! https://github.com/grafana/metrics-drilldown/pull/384/commits/dbd65840f2ebfbb272a43c5253ea1ef7740612dd

src/DataTrail.tsx
NWRichmond: Good idea, implemented in https://github.com/grafana/metrics-drilldown/pull/384/commits/7af7247b9e60ddefd40c905038716bf863517cbc

grafakus (approved):
Nice job! As discussed in a call, we'll follow-up with https://github.com/grafana/metrics-drilldown/issues/401 after this PR is merged.

PR #385, opened 2025-05-05
chore: remove homepage code

shelldandy:
Fixes: https://github.com/grafana/metrics-drilldown/issues/255


NWRichmond (approved):
Looks good!

Could you please also remove the `homepage` image from `info.screenshots` in `src/plugin.json`? There are some other homepage-related things in the codebase still, but the homepage screenshot is the only one that is user-facing.

NWRichmond (approved):
Great! Thank you for this, @shelldandy - it's a gift, especially in terms of bundle size.

PR #386, opened 2025-05-05
fix: playright published broad permissions

shelldandy:
Narrow down the permissions from the whole workflow to just the step that needs it


NWRichmond (approved):
Nicely done!

Issue #355, opened 2025-04-28
[Bug bash] Update Y-Axis label logic


shelldandy:
<img width="699" alt="Image" src="https://github.com/user-attachments/assets/45807d8d-f93a-4b1b-8bd3-c90659d54537" />

PR #390, opened 2025-05-06
fix(Units): Change default unit from "short" to "none"

grafakus:
### ✨ Description

**Related issue(s):** fixes https://github.com/grafana/metrics-drilldown/issues/355

This PR changes the default unit that is used for all viz panels to prevent shortening the numbers on the y axis, and ending up in the situation described in the related issue. 

### 📖 Summary of the changes

The change is trivial, the other modifications in this PR are to update some unit test assertions and regenerate the E2E tests screenshots.

### 🧪 How to test?

- The build should pass


src/autoQuery/units.ts
```diff
@@ -1,4 +1,4 @@
-export const DEFAULT_UNIT = 'short';
+export const DEFAULT_UNIT = 'none';
```
grafakus, line 1: This is the change :)

PR #376, opened 2025-05-01
ci(*): Add ESLint plugin for SonarJS

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR adds the [ESLint plugin for SonarJS](https://www.npmjs.com/package/eslint-plugin-sonarjs) to our app in order to automate some good practices and reduce the complexity of our code base. It's a starting point, we can choose to agree on some rules and to refuse or tweak others. Let's talk!

### 📖 Summary of the changes

Each commit to the PR addresses a specific rule. See diff tab for specific comments.

### 🧪 How to test?

- The build should pass

PR #394, opened 2025-05-08
fix: retry until assertion passes

NWRichmond:
### ✨ Description

A change [here](https://github.com/grafana/metrics-drilldown/pull/376/files#diff-889ff398a9844294fd45acd689975a41466b984ef8b124b1cae2013f159126f7) caused an e2e test to be flaky.

### 📖 Summary of the changes

This PR restores the `expect(...).toPass()` wrapper around the usage count assertion. This tells Playwright to keep trying the assertion until is passes, which avoids erroneous failures caused by loading behavior. For context on the `expect(...).toPass()` pattern, see https://playwright.dev/docs/test-assertions#expecttopass.

### 🧪 How to test?

All CI should pass.


yangkb09 (approved):
yay!!

PR #397, opened 2025-05-09
fix(Histograms): Increase histogram panel height to prevent cropping

grafakus:
### ✨ Description

**Related issue(s):** `-`

The histogram panel legends are cropped. This PR increases the height of all the panels around the app to prevent this to occur (+20px).

| Before | After |
|  ---   |  ---  |
| image | image |
| image |image|
| image | image |

### 📖 Summary of the changes

The change is trivial. See diff tab

### 🧪 How to test?

- The build with the updated screenshots should pass


catherineymgui (approved):

yangkb09 (approved):
so good - thank you!!

PR #398, opened 2025-05-09
fix(VizPanels): Move the "Native Histogram" badge before the panel title

grafakus:
### ✨ Description

**Related issue(s):** is blocked by https://github.com/grafana/metrics-drilldown/pull/397

To prevent confusion with the panels CTAs, this PR moves the "Native Histogram" badge before the panel title.

| Before | After |
|  ---   |  ---  |
| image | image |

### 📖 Summary of the changes

A little CSS hack, see diff tab

### 🧪 How to test?

- The build should pass


src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx
```diff
@@ -18,27 +18,29 @@ import { trailDS } from 'shared';
...
   prometheusFunction?: PrometheusFn;
   title?: string;
 }
 
 interface MetricVizPanelState
   extends SceneObjectState,
-    Pick<Required<MetricVizPanelProps>, 'height' | 'highlight' | 'prometheusFunction'> {
+    Pick<
+      Required<MetricVizPanelProps>,
+      'isNativeHistogram' | 'headerActions' | 'height' | 'hideLegend' | 'highlight' | 'prometheusFunction'
```
grafakus, line 42: It starts to be unreadable... :/

catherineymgui:
Hmm taking a look at this now, I feel like since the native histogram badge is quite long - now that I'm seeing it on the left of the title, I feel like it pushes the title a bit too much and makes it slightly harder to read in comparison to the two badges side-by-side. What are your thoughts?

yangkb09:
yeah i agree with catherine - now that i'm seeing it implemented, it feels a bit unnatural to have the metric name be pushed towards the middle of the panel. i actually prefer having the NH badge next to the select - what do you think, marc?

grafakus:
It's configurable. I've reduced the width of the badge (and its padding):
image

yangkb09:
yes, i tried the PR locally! 

hm, are we able to have the badge say "NH" and then on hover, it says "Native Histogram"? what do you think @grafakus @catherineymgui? even with the reduced padding, etc the badge still appears rather wide when place before the metric title and disrupts the flow/decreases ability to skim things quickly imo

either way i dont think it's too big of a deal and am happy to approve this change if we just want to get it in and iterate on it in the future

grafakus:
> hm, are we able to have the badge say "NH" and then on hover, it says "Native Histogram"?

We would need some JS to display a tooltip... Maybe just a panel description would be enough:
image

Or is it critical to keep the badge?

catherineymgui:
Yeah, I would confirm with Zhehao to determine how long we should keep the badge for!

I don't think it's necessary to have two elements that describe the native histograms, I do like Kat's suggestion of the badge with a tooltip over it. 

I also know the panel description info icon has also been used for the metric description and type, right?

bohandley (approved):
LGTM!

PR #444, opened 2025-05-15
chore: Remove VSCode settings, update README for VSCode users

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR removes the VSCode settings from the code base and adds them as a suggestion for VSCode users in our README.

Committing VSCode settings to a repository is generally not recommended for several reasons:
1. **Personal preferences** - Editor settings are often tailored to individual developers' preferences
2. **Machine-specific configurations** - Some settings may include absolute paths or hardware-specific optimizations
3. **Team diversity** - Team members may use different editors or IDEs
4. **Repository bloat** - Adds unnecessary files to the repository
5. **Security risks** - Settings might contain sensitive information like API keys or tokens
6. **Conflicts** - Can lead to frequent merge conflicts as team members update their settings

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

`-`


grafakus:
Mmmmh...

> True, but user settings override settings defined in the project-level `.vscode/settings.json`.

As a contributor, why am I forced to update my user settings just for the metrics drilldown project? 

> Neither of these apply to our .vscode/settings.json. I expect that both of these things should be caught in PR review.

I think we miss the security aspect here: on one hand, we  don't control which extensions our contributors install on their machine and, on the other hand, once they are committed and pushed, they're public in Git history and can be retrieved.

> Considering the extent that our team uses VSCode or editors underpinned by VSCode, I believe this helps us stay consistent.

Indeed, our IDEs, by being configured a certain way, can help our code base to stay consistent, but eventually, our automation is the one that enforces this consistency. What if the team decides to use a different editor? What about external contributors? 🤷🏾‍♂️ 

> But this feels like a configuration issue that's up to us as individuals to avoid or resolve. For example, perhaps the extension in question can be configured to change the User Settings instead of the Workspace Settings. If not, then the rogue extension can be disabled for this project

Why creating this friction in the first place? Why would an open source project force its contributors to configure their personal IDE a certain way or worse, to disable some extensions when they want to contribute to the project?!
As a contributor, I'd expect the project to have all the automation set up to enforce its standards, not to force my IDE choices.

grafakus:
An example where having the VSCode config in the repo does not help: [this build is failing](https://github.com/grafana/metrics-drilldown/actions/runs/15045246736/job/42286051261) for a lint issue that should be fixed automatically whenever a VSCode user saves a file.

yangkb09:
i also don't feel strongly either way - i see both points you two are making @grafakus @NWRichmond

NWRichmond (approved):
> Indeed, our IDEs, by being configured a certain way, can help our code base to stay consistent, but eventually, our automation is the one that enforces this consistency.

Absolutely - I want to be clear that the point of `.vscode/settings.json` wasn't project-wide consistency, but rather "creature comforts" that reduce the number of issues that result in CI failures. With that in mind, I like @bohandley's suggestion of using git hooks (perhaps pre-push?) to check for formatting & linting issues. It's an IDE-agnostic approach that prevents unnecessary CI failures.

To avoid protracted discussion of `.vscode/settings.json`'s merits & shortcomings, I'm approving this PR, and look forward to seeing formatting & linting issues caught via git hooks.

As a final step, perhaps we can add `.vscode/` to `.gitignore`.

PR #454, opened 2025-05-21
Fix: Correctly fire quick search RudderStack event

yangkb09:
### ✨ Description

Ref #304 

Previously, we were over-reporting quick search usage. The old implementation fired the RudderStack event for new keystrokes after a 1 second debounce. 

The new implementation fires after 1) checking to see if the quick search state is empty, and 2) if the new state is not empty. That way the event fires only for each new search value.

### 🧪 How to test?

See that the console.log only fires for an entirely new search
304:quick-search-log

https://github.com/user-attachments/assets/e0bdc505-94a6-4cb9-a580-e033052cba3d


bohandley (approved):
LGTM!

NWRichmond (approved):
Looks good! I suspect we'll lose the "I previously searched for something, and now I'm changing my search term" search events (unless the search is cleared before using the new search term), but that seems preferable to over-reporting. This is also a win in terms of simplicity.

Issue #455, opened 2025-05-22
Error in breakdown view when series are empty


NWRichmond:
## Context

In the Metric scene, when selecting a label in the breakdown view, we see an error when encountering a series with empty `fields`:

```
zone.js:163 Uncaught TypeError: Cannot read properties of undefined (reading 'values')
    at sorting.ts:64:68
```

Example series with empty `fields`:

```json
[
  {
    "fields": [],
    "length": 0
  }
]
```

PR #456, opened 2025-05-22
Breakdown: Fix sorting with empty series fields

NWRichmond:
### ✨ Description

**Related issue(s):** This PR fixes https://github.com/grafana/metrics-drilldown/issues/455.

The goal is to fix an error related to an unhandled case (empty series) when sorting label values in the Metric scene's Breakdown tab.

### 📖 Summary of the changes

1. I wrote an "empty series" test that failed before the fix.
2. A sorting function was updated to handle empty series.

### 🧪 How to test?

- [x] All CI should pass

### DIFF
--- a/src/services/sorting.test.ts
+++ b/src/services/sorting.test.ts
@@ -41,6 +41,7 @@ const frameC = toDataFrame({
     },
   ],
 });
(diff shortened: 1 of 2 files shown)


bohandley (approved):
LGTM!

PR #463, opened 2025-05-27
feat(Tracking): Add app initialized Rudderstack event

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR tracks when the app is initialized so we can update our product KPIs dashboard

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

`-`


bohandley:
We'll need to identify where it is being started from. The previous rudderstack event identified the following:

Event: `grafana_explore_metrics_exploration_started`

cause:
- `bookmark_clicked`:a bookmark was clicked from the home page
- `recent_clicked`: a recent exploration was clicked from the home page
- `new_clicked`: "new exploration" was clicked from the home page
- `loaded_local_recent_url`: the page was loaded (or reloaded) from a URL which matches one of the recent explorations
- `loaded_shared_url`: the page was loaded from a URL which did not match one of the recent explorations, and is assumed shared
- `dashboard_panel`: the exploration was opened from the dashboard panel menu and is embedded in a drawer

We'll eventually need this from Asserts integrations and other entry points :)

src/App/useReportAppInitialized.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (!initialized) {
+      setInitialized(true);
+
+      const view: ViewName = new URL(window.location.href).searchParams.get('metric')
+        ? 'metric-details'
+        : 'metrics-reducer';
+
+      reportExploreMetrics('app_initialized', { view });
+    }
+  }, [initialized]);
```
NWRichmond, line 18: I think a ref would be nice to use here, instead of state:

```suggestion
import { useEffect, useRef } from 'react';

import { reportExploreMetrics, type ViewName } from 'interactions';

export function useReportAppInitialized() {
  const initialized = useRef(false);

  useEffect(() => {
    if (!initialized.current) {
      initialized.current = true;

      const view: ViewName = new URL(window.location.href).searchParams.get('metric')
        ? 'metric-details'
        : 'metrics-reducer';

      reportExploreMetrics('app_initialized', { view });
    }
  }, []);
```

Using a ref for this helps avoid an unnecessary re-render when `setInitialized(true)` is called.

NWRichmond (approved):
Looks good! Just one performance `nit:`

src/App/useReportAppInitialized.ts
grafakus: True, thank you Nick! Here it is: https://github.com/grafana/metrics-drilldown/pull/465

PR #471, opened 2025-06-05
feat: add slack notification to our slack channel

shelldandy:
Lets test out the slack notifications from the CD action


grafakus (changes requested):
Let's create a dedicated channel for the notifications so we don't create noise in our project channel and we keep things clear when we want to understand what's happening with our pipeline. `#proj-metricsdrilldown-ops`? `#proj-metricsdrilldown-cd`? Let's decide with the team during our weekly what's the preferred option.

PR #479, opened 2025-06-11
refactor(QuickSearch): Improve QuickSearch

grafakus:
### ✨ Description

**Related issue(s):** https://github.com/grafana/metrics-drilldown/pull/477

Another refactor to ease the integration into Asserts (there are more to come).

### 📖 Summary of the changes

This refactor makes the `QuickSearch` component easier to reuse by enabling the configuration of the target variables (the variables that are filtered). See diff tab for specific comments.

### 🧪 How to test?

- The build should pass

### DIFF
--- a/src/WingmanDataTrail/MetricsReducer.tsx
+++ b/src/WingmanDataTrail/MetricsReducer.tsx
@@ -66,7 +66,7 @@ export class MetricsReducer extends SceneObjectBase<MetricsReducerState> {
   protected _variableDependency = new VariableDependencyConfig(this, {
     variableNames: [VAR_WINGMAN_GROUP_BY],
     onReferencedVariableValueChanged: (variable) => {
-      this.updateBodyBasedOnGroupBy((variable as LabelsVariable).state.value as string);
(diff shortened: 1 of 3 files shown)


src/WingmanDataTrail/MetricsReducer.tsx
```diff
@@ -88,13 +88,26 @@ export class MetricsReducer extends SceneObjectBase<MetricsReducerState> {
...
+
+    this.updateBasedOnGroupBy(groupByValue);
 
     this.subscribeToEvents();
   }
 
+  private updateBasedOnGroupBy(groupByValue: string) {
+    const hasGroupByValue = Boolean(groupByValue && groupByValue !== NULL_GROUP_BY_VALUE);
+
+    sceneGraph.findByKeyAndType(this, 'quick-search', QuickSearch).toggleCountsDisplay(!hasGroupByValue);
```
grafakus, line 102: Note: we hide the counts in case of group by because each metrics group can have a different count, so it does not make sense to display a unique value in the quick search.

yangkb09 (approved):

PR #487, opened 2025-06-13
Bump version v1.0.2

shelldandy:
### ✨ Description

Prepare for patch version release


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "grafana-metricsdrilldown-app",
-  "version": "1.0.2-0",
+  "version": "1.0.2",
```
yangkb09, line 3: silly Q - would this not be version `1.0.3`? or `1.1.0`? i wasn't familiar with when to increase the minor/patch version numbers but reading [this](https://semver.org/) seems like we'd at least bump the patch number?

NWRichmond (approved):
Looks good!

PR #507, opened 2025-06-19
refactor(*): Remove unused code

grafakus:
### ✨ Description

**Related issue(s):** `-`

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- The build should pass

### DIFF
--- a/src/WingmanDataTrail/Labels/LabelsVariable.tsx
+++ b/src/WingmanDataTrail/Labels/LabelsVariable.tsx
@@ -11,10 +11,6 @@ import { Label, useStyles2 } from '@grafana/ui';
 import React from 'react';
 
 import { VAR_DATASOURCE, VAR_FILTERS_EXPR } from 'shared';
-import {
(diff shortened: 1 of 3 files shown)


src/WingmanDataTrail/Labels/LabelsVariable.tsx
```diff
@@ -39,22 +35,6 @@ export class LabelsVariable extends QueryVariable {
   }
 
   onActivate() {
-    const filteredMetricsVariable = sceneGraph.lookupVariable(
```
grafakus, line 42: This is useless because when the label value changes (see [MetricsReducer](https://github.com/grafana/metrics-drilldown/blob/main/src/WingmanDataTrail/MetricsReducer.tsx#L99-L109)), the `MetricsList` changes to a `MetricsGroupByList`, which eventually creates instances of `MetricsWithLabelValueVariable`.

I think it's a remain of the prototype of the onboarding flow we did during the hackathon week.

shelldandy (approved):

Issue #442, opened 2025-05-14
[spike] Understand how we can add a panel to dashboards


grafakus:
- Considering the different Grafana versions we support
- Existing API? Extension links?
-

bohandley:
Update research with POCs

1. Expose the add to dashboard code from Grafana as a service
  - https://github.com/grafana/grafana/pull/107150
  - Handles auth
  - Handles dashboard schema
  - Can accept any type of panel with visualizations and query options so handles histograms

2. Companion PR in metrics drilldown
  - https://github.com/grafana/metrics-drilldown/pull/510
  - Accesses runtime service from Grafana
  - Passes a panel to the service to add to dashboard

Feedback:

- Exposing the Grafana code as a singleton is not a path that @grafana/plugins-platform wants to support so these were offered as suggestions (at the end of June)
- (1) Expose using the React context - e.g. we add the API to a React context shared with plugins, and then we expose a useAddPanelToDashboard() hook through grafana-runtime, that can be used to access the API function for opening the popup. The downside is that it can only be used in a React context.
- (2) Expose the component and the service separately - The component would be just a component, and the services would be injected to it via props. Any core services used by the component would still be shared with plugins using singletons (I think this would be a less convenient approach for plugin authors).
- (3) Use the Event Bus - we could create a new event and expose it to plugins, and dispatching that event (with the panel as a payload) the add-to-dashboard modal would be opened by core.

bohandley:
Current Progress:

[@grafana/plugins-platform is working on a strategy for restrictive apis here.](https://github.com/grafana/grafana/pull/107967)

The idea is that Grafana APIs that 

> We are usually exposing core services using singletons (setters + getters) in grafana-runtime. The difference here is, that as those are static dependencies, they are available to every plugin, while here we would like to share APIs restrictively only with a predefined set of plugins.

Once this work is completed, I will be able to migrate the POCs to this strategy to access the "add to dashboard" code in Grafana so that metrics drilldown app users can add to dashboard.

Closing this spike issue, as we now know how to add to dashboard.

Issue #423, opened 2025-05-13
[VizPanel] Allow our users to add any panel to a dashboard


grafakus:
**As a** Metrics Drilldown user
**I want to** be able to add any panel to a dashboard 
**So that** it simplifies the creation of useful dashboards for my team and I (no need to write a query and to configure the panel)

**Acceptance criteria**
1. All the panels (timeseries, histograms, ...) in the app have a new menu item "Add to dashboard"
2. Once clicked on the menu item, the user is offered to add the panel to a new dashboard or an existing one
3. Some End-To-End tests are implemented

bohandley:
Closing this as the above PR's have been merged!

Grafana exports the AddToDashboardForm and Metrics drilldown opens it with a modal so a customer can add a panel to a dashboard.

The metrics to measure usage are 

`grafana_explore_metrics_add_to_dashboard_build_panel`
`grafana_explore_metrics_add_to_dashboard_modal_opened`

PR #510, opened 2025-06-24
feat(*): Add to dashboard PoC

bohandley:
### ✨ Description

**Related issue(s):** 
https://github.com/grafana/metrics-drilldown/issues/442 
https://github.com/grafana/metrics-drilldown/issues/423

https://github.com/user-attachments/assets/64a65563-c9ea-4ae5-8fc4-bb2d956a0e9a

Screenshot 2025-06-24 at 12 15 31 PM
Screenshot 2025-06-24 at 12 16 17 PM
Screenshot 2025-06-24 at 12 16 25 PM

This allows a user to "Add to Dashboard." When a user selects a metric, that metric will have a panel (query, options, visualization, etc) and that panel can be added to a new dashboard or existing dashboard.

We use the `ExploreToDashboardPanel` function by exposing it in Grafana runtime to get all the behaviors in the the feature that is used from the Explore query editor.

### 📖 Summary of the changes

1. In the metric scene, we add an additional button that gets the panel data from the main graph and fires the `PanelDataRequestEvent`.
2. `DataTrail.tsx` listens for the event and when the panel data is recieved, opens the model for adding to dashboard.
3. In the modal we use the `AddToDashboardService` which is the same code used in grafana/grafana in the Explore feature to "Add to dashboard."
4. `AddToDashboardService` is created in grafana and accessible via grafana/runtime so that we have access to user permissions through the context server, to dashboard schema, and all the logic in the Explore Add to Dashboard feature.
5. We can add to a new or existing dashboard.

### 🧪 How to test?

1. Run both Grafana and Metrics Drilldown locally
2. Use this `branch` in grafana which exposes the add to dashboard service through grafana runtime
3. In Grafana, use the feature toggle so that you can use your local metric drilldown app. Otherwise, it will preinstall metrics drilldown from the plugin library into your `./plugins` directory.
```
[plugins]
preinstall_disabled = true
```
4. Navigate to metrics drilldown.
5. Select a metric
6. In the metric scene, click the 'Add to dashboard button' which is next to the 'Open in explore' button.
Screenshot 2025-06-24 at 12 15 31 PM
7. Select a new dashboard or add to a previously created dashboard.
8. Confirm that the panel JSON is the same from the drilldown app with the following panel fields (and time range)
  - datasource
  - query
  - visualization
  - query options
  - legend for the panel

Screenshot 2025-06-24 at 12 19 09 PM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,9 @@ jobs:
     with:
       run-playwright: false
       run-playwright-docker: true
-      run-playwright-with-skip-grafana-dev-image: ${{ inputs.skip-grafana-dev-image || false }}
+      # TEMP until the Docker Hub issue is fixed (2025-07-09)
+      run-playwright-with-skip-grafana-dev-image: true
+      # run-playwright-with-skip-grafana-dev-image: ${{ inputs.skip-grafana-dev-image || false }}
       upload-playwright-artifacts: true
       playwright-report-path: e2e/test-reports/
       playwright-docker-compose-file: docker-compose.yaml
--- a/.github/workflows/pr-title-validation.yml
+++ b/.github/workflows/pr-title-validation.yml
@@ -1,0 +1,16 @@
+name: 'PR title validation (Conventional Commits spec)'
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+
+jobs:
+  validate-pr-title:
+    runs-on: ubuntu-latest
+    steps:
+      - name: PR Conventional Commit Validation
+        uses: ytanikin/PRConventionalCommits@b628c5a234cc32513014b7bfdd1e47b532124d98 # v1.3.0
+        with:
+          task_types: '["feat","fix","docs","test","ci","refactor","perf","chore","revert","build"]'
+          add_label: 'true'
+          custom_labels: '{"docs": "type/docs"}'
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -17,6 +17,7 @@ services:
       GF_PLUGINS_PREINSTALL_DISABLED: true
     volumes:
       - ./provisioning/dashboards/dashboardJson:/etc/dashboards
+      - ../logs-drilldown/dist:/var/lib/grafana/plugins/grafana-lokiexplore-app
     extra_hosts:
       # This allows us to connect to other services running on the host machine.
       - 'host.docker.internal:host-gateway'
--- a/e2e/config/playwright.config.common.ts
+++ b/e2e/config/playwright.config.common.ts
@@ -66,9 +66,12 @@ export function config(config: CustomEnvConfig) {
       grafanaAPICredentials: getGrafanaUser(),
       /* Base URL to use in actions like `await page.goto('/')`. */
       baseURL: getGrafanaUrl(),
+      // Record trace only when retrying a test for the first time.
       screenshot: 'only-on-failure',
-      video: 'retain-on-failure',
-      trace: 'retain-on-failure',
+      // Record video only when retrying a test for the first time.
+      video: 'on-first-retry',
+      /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+      trace: 'on-first-retry',
     },
     /* Configure projects for major browsers */
     projects: [
--- a/e2e/docker/Dockerfile.playwright
+++ b/e2e/docker/Dockerfile.playwright
@@ -1,9 +1,9 @@
-FROM mcr.microsoft.com/playwright:v1.53.1
+FROM mcr.microsoft.com/playwright:v1.54.0
 
 WORKDIR /app
 
 # required by the e2e test code
-RUN npm install "@playwright/test@^1.53.1" "dotenv@^16.3.1" "@grafana/plugin-e2e"
+RUN npm install "@playwright/test@^1.54.0" "dotenv@^16.3.1" "@grafana/plugin-e2e"
 
 ENV TZ=Europe/Madrid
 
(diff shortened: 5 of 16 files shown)


src/DataTrail.tsx
```diff
@@ -232,6 +241,12 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
     }
   }
 
+  private _handlePanelDataRequestEvent(evt: PanelDataRequestEvent) {
+    // Store the panel data and trigger the modal to open
+    this.setState({ panelData: evt.payload });
```
grafakus, line 246:
```suggestion
    this.setState({ panelData: evt.payload, isAddToDashboardModalOpen: true });
```

The `useEffect` usage below should not be ;)

src/MetricActionBar.tsx
```diff
@@ -86,6 +100,73 @@ export class MetricActionBar extends SceneObjectBase<MetricActionBarState> {
...
+  public onGetPanelData = () => {
+    try {
+      const panelData = this.getPanelData();
+      console.log('Panel Data for Dashboard Creation:', panelData);
+      
+      // Fire the event with panel data to trigger the modal
+      const trail = getTrailFor(this);
+      trail.publishEvent(new PanelDataRequestEvent(panelData), true);
+    } catch (error) {
+      console.error('Error getting panel data:', error);
```
grafakus, line 162: @bohandley  Let's use [displayError](https://github.com/grafana/metrics-drilldown/blob/main/src/WingmanDataTrail/helpers/displayStatus.ts#L6-L15) to display the toast _and_ log the error so we can see it in our FE O11y logs.

Also, maybe we can also update ESLint to fail the build when we use console methods directly (by doing so, we ensure the usage of the Logger class and proper logging to our FE O11y app). Wdyt?

src/MetricActionBar.tsx
```diff
@@ -86,6 +100,73 @@ export class MetricActionBar extends SceneObjectBase<MetricActionBarState> {
     });
   };
 
+  public getPanelData = (): PanelDataRequestPayload => {
```
grafakus, line 103:
```suggestion
  public getPanelData = (vizPanel): PanelDataRequestPayload => {
```

By modifying this function to accept a panel as argument, we make it easier to reuse in the future, in a different context (e.g. if we add a menu item on the panel itself).

src/MetricActionBar.tsx
bohandley: Absolutely, I agree, I'm up for updating ESLint to fail the build and suggesting the logger instead. Thank you!

bohandley: Update here! No more console usage and suggesting to use `logger` instead. I am using the `no-restricted-syntax` rule over `no-console` because there is a conflict between our config and @grafana/eslint-config.

https://github.com/grafana/metrics-drilldown/pull/520

src/DataTrail.tsx
```diff
@@ -171,6 +183,14 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
   private _addingFilterWithoutReportingInteraction = false;
   private datasourceHelper = new MetricDatasourceHelper(this);
 
+  private _isAddToDashboardServiceAvailable(): boolean {
+    try {
+      return typeof getAddToDashboardService === 'function' && getAddToDashboardService() != null;
+    } catch {
+      return false;
+    }
```
grafakus, line 191: If `getAddToDashboardService` never throws, we could just use:

```suggestion
    return typeof getAddToDashboardService === 'function' && getAddToDashboardService() != null;
```

src/DataTrail.tsx
```diff
@@ -314,6 +344,24 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
 
     return (
       <div className={styles.container}>
+        {/* add modal at the top of the trail here, in case we add a button for all panels */}
+        {model._isAddToDashboardServiceAvailable() && (
```
grafakus, line 348: Should it be called on every render? Or could we just call it once and store its value, at the top level of this file?

My assumption is that, because it comes from `@grafana/runtime`, the availability of `getAddToDashboardService`can be determined directly once after its import.

src/Menu/PanelMenu.tsx
```diff
@@ -80,6 +80,7 @@ export class PanelMenu extends SceneObjectBase<PanelMenuState> implements VizPan
           onClick: () => exploreUrl?.then((url) => url && window.open(url, '_blank')),
           shortcut: 'p x',
         },
+        // this is where we might add the add to dashboard button, directly on the panel
```
grafakus, line 83: Should we create a GitHub issue instead of a comment?

src/MetricActionBar.tsx
```diff
@@ -62,6 +64,16 @@ export const actionViewsDefinitions: ActionViewDefinition[] = [
 interface MetricActionBarState extends SceneObjectState {}
 
 export class MetricActionBar extends SceneObjectBase<MetricActionBarState> {
+  private _isAddToDashboardServiceAvailable(): boolean {
```
grafakus, line 67: Same comments as above

src/MetricActionBar.tsx
```diff
@@ -86,6 +98,29 @@ export class MetricActionBar extends SceneObjectBase<MetricActionBarState> {
...
+  }
+
+  public onGetPanelData = () => {
+    try {
+      const panelData = this.getMetricScenePanelData();
+      // Fire the event with panel data to trigger the modal
+      const trail = getTrailFor(this);
+      trail.publishEvent(new PanelDataRequestEvent(panelData), true);
+    } catch (error) {
+      displayError(error as Error, ['Failed to retrieve panel data. Please check the console for more details.']);
```
grafakus, line 120: From a UX perspective, it's great to describe the issue and offer advice to our users when this error occurs 👍🏾👍🏾

What do you think if we try to provide advice that is actionable for _all_ our users? I mean we shouldn't assume that they know how to check the console and to understand what will be displayed there (as you know it is sometimes challenging for our team to understand the errors happening in the app!).

Could we display instead a "Please try again later" or "Reload the page" (not great though)? I also personally like to add a "Sorry for the inconvenience" message.

PR #517, opened 2025-06-26
fix: Selecting __name__ as a filter breaks queries

shelldandy:
### ✨ Description

Addresses: https://github.com/grafana/support-escalations/issues/14997

### DIFF
--- a/e2e/config/playwright.config.common.ts
+++ b/e2e/config/playwright.config.common.ts
@@ -66,12 +66,9 @@ export function config(config: CustomEnvConfig) {
       grafanaAPICredentials: getGrafanaUser(),
       /* Base URL to use in actions like `await page.goto('/')`. */
       baseURL: getGrafanaUrl(),
-      // Record trace only when retrying a test for the first time.
       screenshot: 'only-on-failure',
-      // Record video only when retrying a test for the first time.
-      video: 'on-first-retry',
-      /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-      trace: 'on-first-retry',
+      video: 'retain-on-failure',
+      trace: 'retain-on-failure',
     },
     /* Configure projects for major browsers */
     projects: [
--- a/e2e/tests/metrics-reducer-view.spec.ts
+++ b/e2e/tests/metrics-reducer-view.spec.ts
@@ -63,7 +63,7 @@ test.describe('Metrics reducer view', () => {
           await metricsReducerView.sidebar.selectGroupByLabel('db_name');
           await metricsReducerView.assertMetricsGroupByList();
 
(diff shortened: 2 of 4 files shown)


bohandley:
The old fix allowed for the following:

1. You can still select `__name__` to filter metrics in the metric select scene, now metrics reducer.
2. The `__name__` filter is not interpolated into the query.

This PR's implementation looks to only remove the label from the adhoc filter. We still want to allow people to select it for the big list of metrics, but we don't want use it in the query :)

src/utils.ts
```diff
@@ -181,7 +181,9 @@ export function limitAdhocProviders(
         opts.queries = [];
       }
 
-      let values = (await datasourceHelper.getTagKeys(opts)).slice(0, MAX_ADHOC_VARIABLE_OPTIONS);
+      let values = (await datasourceHelper.getTagKeys(opts))
```
bohandley, line 184: We just want to remove it from query interpolation, not remove it as a value to select in the adhocFilters variable :)

bohandley (changes requested):

e2e/tests/metrics-reducer-view.spec.ts
```diff
@@ -11,6 +11,32 @@ test.describe('Metrics reducer view', () => {
...
+        .click();
+
+      await metricsReducerView.assertCoreUI();
+      await expect(
+        metricsReducerView
+          .getByTestId('data-testid Panel header a_utf8_http_requests_total')
+          .getByTestId('data-testid Panel status error')
+      ).not.toBeVisible();
+    });
+  });
```
shelldandy, line 38: This is the assertion for the panel error not being there:

image

HOWEVER... for some reason this only triggers when setting playwright to record a video, not on headless mode, hence why i added `retain-on-failure` on the playwright options to force it to register the error, or in this case to force it to register the error no longer happening

e2e/config/playwright.config.common.ts
```diff
@@ -66,12 +66,9 @@ export function config(config: CustomEnvConfig) {
...
       /* Base URL to use in actions like `await page.goto('/')`. */
       baseURL: getGrafanaUrl(),
-      // Record trace only when retrying a test for the first time.
       screenshot: 'only-on-failure',
-      // Record video only when retrying a test for the first time.
-      video: 'on-first-retry',
-      /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-      trace: 'on-first-retry',
+      video: 'retain-on-failure',
+      trace: 'retain-on-failure',
```
shelldandy, line 71: see comment for the test about this

bohandley: That's interesting! Do we do any other e2e tests for failed queries or errors that are returned from Prometheus in a panel?

PR #518, opened 2025-06-26
perf: avoid transitive dependencies from `@grafana/prometheus`

### LINKED ISSUE #410: Reduce the bundle size impact of `@grafana/prometheus`
## Context

Currently, `@grafana/prometheus` defaults to using [`moduleResolution: 'classic'`](https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution). In Grafana Metrics Drilldown, `@grafana/prometheus` is an "all or nothing" import - when we import anything, it imports the whole package, and all of its transitive dependencies come along for the ride. We can't take full advantage of tree-shaking, and this has some unfortunate consequences on our bundle size.

This pain was first felt in https://github.com/grafana/metrics-drilldown/issues/63, and we've had to use some hacks to get around the bundle size impacts. This also limits our ability to expose [component-type extension points](https://grafana.com/developers/plugin-tools/how-to-guides/ui-extensions/create-an-extension-point) for embedding parts of Grafana Metrics Drilldown in other parts of Grafana / apps. We'd like to be able to do that, but it's not practical until `@grafana/prometheus` becomes a tree-shakeable dependency.

## Solution

Create & merge a PR that solves https://github.com/grafana/grafana/issues/99111. I've already created a [branch](https://github.com/grafana/grafana/compare/99111/prom-bundler-mode-update) that does this, but it needs a little bit more work to get to the finish line.

NWRichmond:
### ✨ Description

This PR closes #410.

#### Simplified webpack config

The primary goal of this PR is to upgrade `@grafana/prometheus` to take advantage of https://github.com/grafana/grafana/pull/105575, which will help us keep our bundle size in check. It also allows us to remove the `NormalModuleReplacementPlugin` hacks that we needed to keep transitive dependencies from `@grafana/prometheus` out of our bundle. According to the `analyze` script, the `@grafana/prometheus` is much tinier [than it used to be](https://github.com/grafana/grafana/pull/105575):

demo gmd analyze bundle grafana prometheus

### Refactoring

One side-effect of upgrading `@grafana/prometheus` to take advantage of https://github.com/grafana/grafana/pull/105575 is that we also receive changes from https://github.com/grafana/grafana/pull/105818, which deprecates many `languageProvider` methods. Because we must support `@grafana/prometheus` versions `11.6.0` through the latest `12.x`, this PR updates `src/helpers/MetricDatasourceHelper.ts` to include a couple of new methods that encapsulate all of the complexity around the `@grafana/prometheus` API changes.

### 🧪 How to test?

All CI should pass.


src/WingmanDataTrail/GroupBy/MetricsWithLabelValue/MetricsWithLabelValueDataSource.ts
```diff
@@ -57,12 +52,12 @@ export class MetricsWithLabelValueDataSource extends RuntimeDataSource {
     const removeRules = query.startsWith('removeRules');
     const matcher = removeRules ? query.replace('removeRules', '') : query;
 
-    if (MetricDatasourceHelper.datasourceUsesTimeRangeInLanguageProviderMethods(ds)) {
-      metricsList = await ds.languageProvider.fetchSeriesValuesWithMatch(timeRange, '__name__', matcher);
-    } else {
-      // @ts-expect-error: Ignoring type error due to breaking change in fetchSeriesValuesWithMatch signature
-      metricsList = await ds.languageProvider.fetchSeriesValuesWithMatch('__name__', matcher);
-    }
+    metricsList = await MetricDatasourceHelper.fetchLabelValues({
```
bohandley, line 55: Nice refactor :)

src/helpers/MetricDatasourceHelper.ts
```diff
@@ -48,30 +52,47 @@ export class MetricDatasourceHelper {
...
+
+    const loadMetadata =
+      typeof ds.languageProvider.retrieveMetricsMetadata === 'function'
+        ? () => Promise.resolve(ds.languageProvider.retrieveMetricsMetadata())
+        : () => (ds.languageProvider.loadMetricsMetadata as unknown as Promise<void>).then(queryMetadata); // eslint-disable-line sonarjs/deprecation
+
+    let metadata = await queryMetadata();
+
+    if (!metadata) {
+      metadata = await loadMetadata();
```
bohandley, line 83: Good coverage for both cases, old and new

bohandley (commented):
I found a weird bug which I can't find the cause for. Group by labels is passing the index, not the name of the label now 🤔 

Screenshot 2025-06-26 at 12 58 03 PM

bohandley:
Found it the Group by bug! The fetch labels response is different [here](https://github.com/grafana/metrics-drilldown/blob/8c69b1145e269252cd373f2454e8e6bd93ea2963/src/WingmanDataTrail/Labels/LabelsDataSource.ts#L110-L118). We'll do the old switcheroo from keys to values for the Group by.

https://github.com/grafana/metrics-drilldown/blob/8c69b1145e269252cd373f2454e8e6bd93ea2963/src/WingmanDataTrail/Labels/LabelsDataSource.ts#L110-L118

New version
Screenshot 2025-06-26 at 1 03 31 PM

Old version 
Screenshot 2025-06-26 at 1 07 03 PM

shelldandy (approved):
seems DRYer than before, love that

NWRichmond:
Thanks so much for the reviews! The **Group by** functionality has been restored.

bohandley (changes requested):
Found another bug! 🦋  In Group by labels, we make a call for metrics for each label, but in this call, we are not interpolating ${filters}, so no metrics are being returned.

https://github.com/grafana/metrics-drilldown/blob/734a1a100bf82bad51d33971e0bcab6867351f86/src/WingmanDataTrail/GroupBy/MetricsWithLabelValue/MetricsWithLabelValueDataSource.ts#L55-L60

Screenshot 2025-06-26 at 3 47 36 PM

Screenshot 2025-06-26 at 3 49 27 PM

bohandley:
> Hi @bohandley! I just rebased, and I can't replicate the issue in your screenshot with (what I presume is) the same TSDB, same label, and same label search term. Do you have any suggestions about how to reproduce this?


I'm running grafana locally so there may be a bug in that? Git pulling and testing it with the most up to date grafana now...

bohandley:
That was it! Sorry about that, I just needed to update Grafana 🤦🏻

bohandley (approved):
LGTM! Great work on this! 🚀

PR #519, opened 2025-06-27
test(PrometheusQuery): Add __name__ unit tests

shelldandy:
Follow up on: https://github.com/grafana/metrics-drilldown/pull/517/

- Moving e2e config back to video record only on fail
- Making a test `async` that was missing it
- Unit test for the `buildPrometheusQuery` filter function

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -17,6 +17,7 @@ services:
       GF_PLUGINS_PREINSTALL_DISABLED: true
     volumes:
       - ./provisioning/dashboards/dashboardJson:/etc/dashboards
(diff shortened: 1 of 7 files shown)


bohandley (approved):
LGTM! 🚀

grafakus (changes requested):
Thank you for the follow-up Miguel!

As we discussed during our last call, the idea was to substitute the E2E test (more costly) for the unit test, which would give us confidence that the Prometheus query is always correct, even when a `__name__` filter is present.

We can remove the new E2E completely.

grafakus (approved):
👍🏾👍🏾

PR #529, opened 2025-07-08
chore: Update @grafana/prometheus dependency and fix a type issue

itsmylife:
### ✨ Description

For the sake of consistency new language provider API was using empty matchers (`{__name__!=""}`) but this is causing some overheat on the database (mimir, prometheus, etc) side. I made some changes to not use empty matchers. See here: https://github.com/grafana/grafana/pull/107569

This PR is bringing this change in metrics-drilldown too. 

Also I noticed there is a typescript error, so I fixed it too. 

### 📖 Summary of the changes

- Bump `@grafana/prometheus` dependency
- Fix type error in `MetricDatasourceHelper.ts`

### 🧪 How to test?

While using a datasource which has support labels API in metrics-drilldown app, we should see labels API calls without empty matchers. (`{__name__!=""}`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
         "@grafana/schema": "^12.0.0",
         "@grafana/ui": "^12.0.0",
         "@lezer/common": "^1.2.3",
+        "@prometheus-io/lezer-promql": "^0.305.0-rc.1",
         "p-limit": "^6.2.0",
         "react": "18.3.1",
         "react-dom": "18.3.1",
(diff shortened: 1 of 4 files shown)


itsmylife:
I don't know why e2e tests are failing.

NWRichmond:
> I don't know why e2e tests are failing.

There's currently a Docker Registry outage that's affecting e2e tests. Once https://www.dockerstatus.com is clear, the tests should pass :)

src/helpers/MetricDatasourceHelper.ts
```diff
@@ -75,7 +75,7 @@ export class MetricDatasourceHelper {
     const loadMetadata =
       typeof ds.languageProvider.retrieveMetricsMetadata === 'function'
         ? () => Promise.resolve(ds.languageProvider.retrieveMetricsMetadata())
-        : () => (ds.languageProvider.loadMetricsMetadata() as unknown as Promise<void>).then(queryMetadata); // eslint-disable-line sonarjs/deprecation
+        : () => (ds.languageProvider.loadMetricsMetadata() as unknown as Promise<void>).then(() => queryMetadata()); // eslint-disable-line sonarjs/deprecation
```
shelldandy, line 78: why would this be different?

itsmylife: @grafakus We cannot because `loadMetricsMetadata` is marked for deprecation. When metrics-drilldown stop supporting old version of grafana you can delete and only use new methods. 

@shelldandy Not sure but I think `queryMetadata` was expecting an optional limit parameter. Typescript didn't like the previous version I made this.

Issue #523, opened 2025-07-05
[MetricScene] Bug: Labels from explore link not respected


bohandley:
### What is this? 🐛 

When linking from grafana explore to metrics drilldown, there may be labels in the url. 

URL from Explore with `var-filters`

`/a/grafana-metricsdrilldown-app/drilldown?metric=app_ads_ad_requests_total&var-ds=fenlcp4kzo4jke&var-filters=deployment_environment%3Doteldemo03`

But, when this link is navigated to, the `var-filters` are removed.

`/a/grafana-metricsdrilldown-app/drilldown?metric=app_ads_ad_requests_total&var-ds=fenlcp4kzo4jke&var-filters=&nativeHistogramMetric=&from=now-1h&to=now&timezone=browser&var-groupby=$__all&actionView=breakdown&layout=grid`

Example:
<img width="576" height="323" alt="Image" src="https://github.com/user-attachments/assets/26fc4ec2-effb-4b6a-bdcd-e42163fdc68c" />

<img width="505" height="315" alt="Image" src="https://github.com/user-attachments/assets/a27b4bee-2efe-431e-aaf3-18a6e8fdb67a" />

PR #540, opened 2025-07-10
bug: Labels require a delimiter when building links

bohandley:
Fixes https://github.com/grafana/metrics-drilldown/issues/523

### ✨ Description

**Related issue(s):** 
https://github.com/grafana/metrics-drilldown/issues/523

Screenshot 2025-07-10 at 6 30 19 PM

Screenshot 2025-07-10 at 6 24 32 PM

When adding labels, the url contains the labels in the `var-filters` param and uses a `|` delimiter between the name, op and value.

`var-filter=service_name|=|adservice`
escaped: `var-filters=service_name%7C%3D%7Cadservice`

We need to add the `|` when building a link so that the explore link and dashboard links respect the label filters when navigating to metrics drilldown.

### 📖 Summary of the changes

Add a `|` delimiter to label filter elements when building a link.

#### Explore link

Screenshot 2025-07-10 at 6 27 47 PM

1. Navigate to Explore and chose a metrics and labels
2. See the go queryless button.
3. Use it to navigate to Metrics drilldown.
4. See labels are respected

#### Dashboard link

Screenshot 2025-07-10 at 6 29 05 PM

1. Open a dashboard with a Prometheus query used in a panel
2. Make sure the Dashboard is NOT in edit mode
3. Click the `...` menu and see metrics drilldown link
4. Use the link to navigate to metrics drilldown and see labels are respected.

### DIFF
--- a/src/extensions/links.test.ts
+++ b/src/extensions/links.test.ts
@@ -1,4 +1,4 @@
-import { parsePromQueryRegex } from './links';
+import { ADHOC_URL_DELIMITER, parsePromQueryRegex } from './links';
 
 describe('parsePromQueryRegex', () => {
   test('should parse basic metric name', () => {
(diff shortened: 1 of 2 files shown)


src/extensions/links.test.ts
```diff
@@ -79,3 +79,28 @@ describe('parsePromQueryRegex', () => {
     expect(result.labelFilters).toEqual([]);
   });
 });
+
+describe('URL Parameter Label Filter Formatting', () => {
```
grafakus, line 83: 🤔 This test does not cover any application code, so how confident can we be that the links are built properly by the application?

What do you think about starting to test the [appendUrlParameters function](https://github.com/grafana/metrics-drilldown/blob/main/src/extensions/links.ts#L93-L106) instead?

By doing so, if a bug is introduced later, the new tests would catch it immediately.

grafakus (changes requested):
🤔 This test does not cover any application code, so how confident can we be that the links are built properly by the application?

What do you think about starting to test the [appendUrlParameters function](https://github.com/grafana/metrics-drilldown/blob/main/src/extensions/links.ts#L93-L106) instead?

By doing so, if a bug is introduced later, the new tests would catch it immediately.

bohandley:
Closing this in favor of https://github.com/grafana/metrics-drilldown/pull/549 because I pushed the change with package-lock, it was easier to start over, I had to refactor `buildNavigateToMetricsParams` to be used in both links (to add pipes), and address test change requests from this PR. Thank you!

Issue #499, opened 2025-01-23
ExploreMetrics: Poor performance in Breakdown scene when dealing with many label values


NWRichmond:
In Explore Metrics, we currently experience unacceptably performance in the Breakdown scene when a label contains many values. The discussion [here](https://raintank-corp.slack.com/archives/C075BDBTX96/p1737645572488119) points to an example where over 2,500 panels are rendered when a certain label is selected.

cakennedy:
@grafakus Given that the PR is In Review, should this Issue also be In Review, not Done?

PR #556, opened 2025-07-17
refactor(MetricScene): Introduce the new GroupByVariable

grafakus:
### ✨ Description

**Related issue(s):** relates to https://github.com/grafana/metrics-drilldown/issues/499

A refactoring PR to prepare for the performance improvement to mitigate the problem described in the linked issue.

### 📖 Summary of the changes

This PR introduces the new `GroupByVariable` in order to split responsibilities and thus, simplify the `LabelBreakdownScene` component.

Two minor "breaking changes":

1. After this PR, we don't track the same Rudderstack events when the user selects a label in the breakdown dropdown vs when the user clicks on a label panel. Instead, we track the new `groupby_label_changed` and `breakdown_panel_selected` events. This should still give us the ability to understand our users journey.
2. Prior to this PR, whenever the ad hoc filters would change, the list of options in the dropdown would be updated to filter out the labels used in the filters. This is not longer the case. It's been done:
  - to simplify the implementation: no need for a dedicated UI for the dropdown, we can rely on the `<QueryVariable.Component />` to handle loading and error states in the UI,
  - because the previous implementation felt wrong: what if the ad hoc filters contain a regex that produces more than a single label value? Shouldn't our users be able to select the corresponding label to drilldown further into its new values?

### 🧪 How to test?

- The build should pass


e2e/fixtures/views/MetricSceneView.ts
```diff
@@ -126,7 +126,7 @@ export class MetricSceneView extends DrilldownView {
   }
 
   async assertLabelDropdown(optionLabel: string) {
-    await expect(this.getLabelDropdown().locator('input')).toHaveValue(optionLabel);
+    await expect(this.getLabelDropdown().getByText(optionLabel).first()).toBeVisible();
```
grafakus, line 129: Caused by the change of markup introduced by the new `GroupByVariable`

src/Breakdown/LabelBreakdownScene.tsx
```diff
@@ -12,63 +12,47 @@ import {
...
-  labels: Array<SelectableValue<string>>;
-  value?: string;
-  loading?: boolean;
-  error?: string;
-  blockingMessage?: string;
 }
 
 export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneState> {
-  protected _variableDependency = new VariableDependencyConfig(this, {
-    variableNames: [VAR_FILTERS],
```
grafakus, line 64: This is now the responsibility of `GroupByVariable`

src/Breakdown/LabelBreakdownScene.tsx
```diff
@@ -122,7 +95,15 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
       this.clearBreakdownPanelAxisValues();
     });
 
-    this.updateBody(variable);
+    const groupByVariable = this.getVariable();
+
+    groupByVariable.subscribeToState((newState, oldState) => {
+      if (newState.value !== oldState.value) {
```
grafakus, line 101: Only when the value changes, the body should be updated. `MetricLabelsList` can handle the loading state and `ByFrameRepeater` as well

src/Breakdown/LabelBreakdownScene.tsx
```diff
@@ -178,12 +159,11 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
   }
 
   private getVariable(): QueryVariable {
-    const variable = sceneGraph.lookupVariable(VAR_GROUP_BY, this)!;
-    if (!isQueryVariable(variable)) {
+    const groupByVariable = sceneGraph.lookupVariable(VAR_GROUP_BY, this)!;
```
grafakus, line 162: Sometimes, naming is easy :p

src/Breakdown/LabelBreakdownScene.tsx
```diff
@@ -203,92 +183,59 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
...
-    const options = getLabelOptions(this, variable);
-
-    const stateUpdate: Partial<LabelBreakdownSceneState> = {
-      loading: variable.state.loading,
-      value: String(variable.state.value),
-      labels: options,
-      error: variable.state.error,
-      blockingMessage: undefined,
-    };
+  private updateBody(groupByVariable: QueryVariable) {
```
grafakus, line 186: Now that the variable has been externalized and handles its own loading state, etc. We don't need to keep track of the variable state here.

src/Breakdown/LabelBreakdownScene.tsx
```diff
@@ -203,92 +183,59 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
...
-    this.setState(stateUpdate);
+        : buildNormalLayout(this._query!, this.onBreakdownLayoutChange, this.state.search),
+    });
   }
 
   public onBreakdownLayoutChange = () => {
     this.clearBreakdownPanelAxisValues();
   };
 
-  public onChange = (value?: string) => {
```
grafakus, line 243: Same here, the variable is responsible for handling what's happening when its value changes (tracking).

src/Breakdown/SelectLabelAction.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+interface SelectLabelActionState extends SceneObjectState {
+  label: string;
+}
+
+export class SelectLabelAction extends SceneObjectBase<SelectLabelActionState> {
+  public onClick = () => {
+    const { label } = this.state;
+
+    reportExploreMetrics('breakdown_panel_selected', { label });
```
grafakus, line 19: See my comment in the PR description.

src/Breakdown/utils.ts
```diff
@@ -1,38 +1,9 @@
...
-    return [];
-  }
-
-  const filters = labelFilters.state.filters;
-
-  for (const option of variable.getOptionsForSelect()) {
-    const filterExists = filters.find((f) => f.key === option.value);
-
-    if (option.label === 'le') {
-      // Do not show the "le" label
```
grafakus, line 21: I can read the code and read the comment, but there's no reason given. 🤷🏾‍♂️
Does anyone know why we would want to remove the "le" label?

src/Breakdown/utils.ts
```diff
@@ -1,38 +1,9 @@
...
-      continue;
-    }
-    if (filterExists) {
-      continue;
-    }
-    labelOptions.push({ label: option.label, value: String(option.value) });
```
grafakus, line 27: "Breaking change": see my comment in the PR description.

src/Breakdown/AddToFiltersGraphAction.tsx
```diff
@@ -13,7 +15,7 @@ export interface AddToFiltersGraphActionState extends SceneObjectState {
 
 export class AddToFiltersGraphAction extends SceneObjectBase<AddToFiltersGraphActionState> {
   public onClick = () => {
-    const variable = sceneGraph.lookupVariable('filters', this);
+    const variable = sceneGraph.lookupVariable(VAR_FILTERS, this);
```
grafakus, line 18: Not related to this PR

src/utils.ts
```diff
@@ -126,7 +126,7 @@ export type SceneTimeRangeState = SceneObjectState & {
 };
 
 export function getFilters(scene: SceneObject) {
-  const filters = sceneGraph.lookupVariable('filters', scene);
+  const filters = sceneGraph.lookupVariable(VAR_FILTERS, scene);
```
grafakus, line 129: Not related to this PR

src/Breakdown/LabelBreakdownScene.tsx
```diff
@@ -312,6 +259,7 @@ function getStyles(theme: GrafanaTheme2) {
       alignItems: 'flex-end',
       gap: theme.spacing(2),
       justifyContent: 'space-between',
+      height: '70px',
```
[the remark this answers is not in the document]
yangkb09: nice

PR #558, opened 2025-07-18
feat: expose embedded label breakdown component

### LINKED ISSUE #515: Convert basic exposed component into breakdown view
## Overview

Following #514, we must update the exposed component to display a label breakdown view.

The breakdown view will inherit the following traits from the Explore query in the RCA Workbench:
- The metric (e.g. `asserts:latency:average`, `asserts:feo11y_error:ratio_5m`, etc.)
- The labels (e.g. `asserts_request_type="inbound"`, `service="frontend"`)

The metric and labels can be passed to our exposed component as props.

## Differences between the embedded & regular breakdown experience

The embedded breakdown experience:
- Instead of a "Select new metric" button, present a button for opening the full Metrics Drilldown app experience
- Must have styles (especially background colors) match the Asserts aoo's RCA Workbench drawer

NWRichmond:
### ✨ Description

**Related issue(s):** Closes #515.

This PR adds a new [exposed component](https://grafana.com/developers/plugin-tools/how-to-guides/ui-extensions/expose-a-component) that other app plugins (in particular, the Asserts app) can use. While it could be used outside of an embedded context, the primary goal is to enable the Label Breakdown experience that Metrics Drilldown users are used to, embedded in another app's UI. For example:

demo embedded label breakdown

### 📖 Summary of the changes

- Configures the new exposed component in [`module.tsx`](https://grafana.com/developers/plugin-tools/how-to-guides/ui-extensions/expose-a-component#expose-a-component-from-an-app-plugin) and [`plugin.json`](https://grafana.com/developers/plugin-tools/reference/plugin-json#extensionsexposedcomponents) as per the docs
- Upgrades `@grafana/scenes` to take advantage of the variable namespacing capabilities introduced in https://github.com/grafana/scenes/pull/1156
- Updates how styling works for `stickyMainGraph`, to allow the **Always keep selected metric graph in-view** functionality to work in embedded mode

### 🧪 How to test?

Please follow the **Testing notes** described in https://github.com/grafana/asserts-app-plugin/pull/1746.

Note that any time range settings in the RCA Workbench should carry over to Metrics Drilldown, but time range changes made in the embedded Metrics Drilldown experience shouldn't change the time range in RCA Workbench. This is consistent with how the embedded Logs and Traces experiences work.


package.json
```diff
@@ -32,10 +32,11 @@
     "@emotion/css": "11.13.5",
     "@grafana/data": "^12.0.0",
     "@grafana/faro-web-sdk": "^1.17.2",
+    "@grafana/i18n": "^12.1.0-249189",
```
NWRichmond, line 35: This is a requirement of the `@grafana/scenes` upgrade, as per [this message](https://raintank-corp.slack.com/archives/C04FRQ2S0AV/p1750937025713649)

src/DataTrail.tsx
```diff
@@ -268,7 +270,7 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
 
         Object.assign(stateUpdate, this.getSceneUpdatesForNewMetricValue(values.metric, nativeHistogramMetric));
       }
-    } else if (values.metric == null) {
+    } else if (values.metric == null && !this.state.embedded) {
```
NWRichmond, line 273: This ensures that when `DataTrail` is initialized with a `metric` in `embedded` mode, the top scene isn't set to `MetricsReducer` when the URL sync happens.

src/DataTrail.tsx
```diff
@@ -302,7 +304,8 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
     const { controls, topScene, settings, pluginInfo, embedded } = model.useState();
 
     const chromeHeaderHeight = useChromeHeaderHeight();
-    const styles = useStyles2(getStyles, embedded ? 0 : chromeHeaderHeight ?? 0);
+    const headerHeight = embedded ? 0 : chromeHeaderHeight ?? 0;
```
shelldandy, line 307:
```suggestion
    const chromeHeaderHeight = embedded ? 0 : useChromeHeaderHeight();
```

shelldandy: first time using the suggestion feature but thats the idea (in case it keeps both lines which is not the suggestion)

src/MetricGraphScene.tsx
```diff
@@ -68,11 +70,11 @@ function getStyles(theme: GrafanaTheme2, chromeHeaderHeight: number) {
     sticky: css({
       display: 'flex',
       flexDirection: 'row',
-      background: theme.isLight ? theme.colors.background.primary : theme.colors.background.canvas,
+      background: getAppBackgroundColor(theme, trail),
       position: 'sticky',
       paddingTop: theme.spacing(1),
       marginTop: `-${theme.spacing(1)}`,
-      top: `${chromeHeaderHeight + 70}px`,
+      top: `calc(var(--app-controls-height, 0px) + ${headerHeight}px)`,
```
shelldandy, line 77: Maybe a comment of how this gets its data would be nice for future readers since it comes from a different file

shelldandy (dismissed):
minor nits which can be followed up on another PR

grafakus (dismissed):
Very nice job Nick! 

One non-blocking suggestion: define `useEmbeddedStyles2(getStyles, model)` to encapsulate the background theme tweak (or any other) in one place instead of many.

src/DataTrail.tsx
NWRichmond: +1, I think we need to call `useChromeHeaderHeight()` non-conditionally to prevent React errors

To avoid the nested ternary, we could do this:
```tsx
const chromeHeaderHeight = useChromeHeaderHeight() ?? 0;
const headerHeight = embedded ? 0 : chromeHeaderHeight;
```

src/DataTrail.tsx
```diff
@@ -312,6 +315,28 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
       limitAdhocProviders(model, filtersVariable, datasourceHelper);
     }, [model]);
 
+    // Set CSS custom property for app-controls height in embedded mode
+    useEffect(() => {
+      // Update on mount and when controls change
```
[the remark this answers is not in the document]
NWRichmond: I tried it, but unfortunately it leads to style issues when not running in embedded mode

package.json
```diff
@@ -30,21 +30,22 @@
   "dependencies": {
     "@bsull/augurs": "^0.9.0",
     "@emotion/css": "11.13.5",
-    "@grafana/data": "^12.0.0",
+    "@grafana/data": "^12.1.0",
```
NWRichmond, line 33: `@grafana` package upgrades are apparently required as per the Compatibility Check action - this is a coincidence and not particular to this PR, but we may as well do them here :)

NWRichmond:
> define `useEmbeddedStyles2(getStyles, model)` to encapsulate the background theme tweak (or any other) in one place instead of many.

I don't understand yet, but I'd like to. Because the `getStyles` functions that use a `DataTrail` have different signatures, I'm not sure how to implement a `useEmbeddedStyles2` hook that would make this cleaner than using `useStyles2`.

src/MetricGraphScene.tsx
NWRichmond: I like that, done in https://github.com/grafana/metrics-drilldown/pull/558/commits/84fbd0f2f2d232c3023f69974fef4a353589e926 ✅

src/DataTrail.tsx
NWRichmond: done in https://github.com/grafana/metrics-drilldown/pull/558/commits/955bf3cb14bc68a02ca2237b883f5e5c2e39fbbc

src/DataTrail.tsx
```diff
@@ -363,29 +393,35 @@ function getVariableSet(initialDS?: string, metric?: string, initialFilters?: Ad
           // remove any filters that include __name__ key in the expression
           // to prevent the metric name from being set twice in the query and causing an error.
           const filtersWithoutMetricName = filters.filter((filter) => filter.key !== '__name__');
-          return [...getBaseFiltersForMetric(metric), ...filtersWithoutMetricName]
+          const filtered = [...filtersWithoutMetricName]
```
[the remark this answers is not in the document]
NWRichmond: Ah yeah, `filtersWithoutMetricName` was made a separate variable for debugging purposes, but we should be able to get by without it.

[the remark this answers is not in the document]
NWRichmond: Done in https://github.com/grafana/metrics-drilldown/pull/558/commits/f7cc46ead9babb159bfb8bf22dfd4f8eb850ce74 ✅

shelldandy (approved):

PR #557, opened 2025-07-18
perf(MetricScene): Add pagination after selecting a label (+ various refactorings)

grafakus:
### ✨ Description

**Related issue(s):** mitigates the issue described in https://github.com/grafana/metrics-drilldown/issues/499, is blocked by https://github.com/grafana/metrics-drilldown/pull/563

In order to prevent the UI to freeze when selecting a high cardinality label on the `MetricScene`, this PR introduces label values pagination, thanks to a customized [SceneByFrameRepeater](https://github.com/grafana/scenes/blob/main/packages/scenes/src/components/SceneByFrameRepeater.tsx) component. 

### 📖 Summary of the changes

The main changes consist in:
1. adding a new customized `SceneByFrameRepeater` component,
2. splitting the `LabelBreakdownScene` into smaller and focused components, to reduce code complexity (mainly by adding a new `MetricLabelValuesList` component, which in turn, uses `SceneByFrameRepeater`),
3. reusing the `QuickSearch` component to filter label values (and adding a counts provider for better UI feedback, see https://github.com/grafana/metrics-drilldown/pull/560),
4. reusing the `LayoutSwitcher`, with custom options and URL search parameter (see https://github.com/grafana/metrics-drilldown/pull/559),
5. cleaning unused files.

An interesting challenge has been to modify the `syncYAxis` behaviour to make it work with pagination, filtering and sorting. Happy to share more details in a call if it's easier for the reviewers :) 

### 🧪 How to test?

- The build should pass

PR #560, opened 2025-07-21
refactor(QuickSearch): Decouple QuickSearch from metric variables

grafakus:
### ✨ Description

**Related issue(s):** made to simplify https://github.com/grafana/metrics-drilldown/pull/557 

A refactor to decouple the [QuickSearch component](https://github.com/grafana/metrics-drilldown/blob/main/src/WingmanDataTrail/ListControls/QuickSearch/QuickSearch.tsx) from the metric variables. Indeed, prior to this PR, the search counts displayed in the UI were directly retrieved from the `MetricsVariable` and the `FilteredMetricsVariable` Scene objects.

This PR will enable us to reuse the `QuickSearch` component in  the "Breakdown" tab of the `MetricScene` and to display the number of labels.

### 📖 Summary of the changes

We introduce the new `CountsProvider` class to be able to customize how counts are retrieved by `QuickSearch`.

### 🧪 How to test?

- The build should pass

### DIFF
--- a/src/WingmanDataTrail/ListControls/QuickSearch/CountsProvider/CountsProvider.ts
+++ b/src/WingmanDataTrail/ListControls/QuickSearch/CountsProvider/CountsProvider.ts
@@ -13,7 +13,6 @@ export class CountsProvider<T extends CountsProviderState = CountsProviderState>
   constructor(state: Partial<T>) {
     super({
       ...state,
-      key: state.key || 'CountsProvider',
       counts: { current: 0, total: 0 },
     } as T);
   }
--- a/src/WingmanDataTrail/ListControls/QuickSearch/CountsProvider/CountsProvider.ts
+++ b/src/WingmanDataTrail/ListControls/QuickSearch/CountsProvider/CountsProvider.ts
@@ -1,0 +1,24 @@
+import { SceneObjectBase, type SceneObjectState } from '@grafana/scenes';
+
+export type CountsData = {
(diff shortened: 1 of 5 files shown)


src/WingmanDataTrail/ListControls/QuickSearch/CountsProvider/CountsProvider.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface CountsProviderState extends SceneObjectState {
+  counts: CountsData;
+}
+
+export class CountsProvider<T extends CountsProviderState = CountsProviderState> extends SceneObjectBase<T> {
+  constructor(state: Partial<T>) {
+    super({
+      ...state,
+      key: state.key || 'CountsProvider',
```
NWRichmond, line 16: Because we're instantiating the `MetricVariableCountsProvider` in a couple of places and using this `state.key || 'CountsProvider'`, I suggest that we either:
- Remove the `|| 'CountsProvider'` so that [`@grafana/scenes` can create a UUID](https://github.com/grafana/scenes/blob/d11f6e4728b5829a5b6121bc555d7e71588abb31/packages/scenes/src/core/SceneObjectBase.tsx#L44-L46), or
- Specify keys in each `new MetricVariableCountsProvider()`

NWRichmond (commented):
Looks good! Just one suggestion:

PR #552, opened 2025-07-16
fix(SeriesSort): Prevent the "Initialize outlier detector first" error

grafakus:
### ✨ Description

**Related issue(s):** loosely relates to https://github.com/grafana/metrics-drilldown/issues/499

Note: this PR is part of many a series of refactoring PRs aimed at improving the performance of the MetricScene (see https://github.com/grafana/metrics-drilldown/issues/499), as well as making its code more maintainable.

The goals of this PR are:
1. Fixing the "Initialize outlier detector first" errors that [have been on the rise for at least a week](https://ops.grafana-ops.net/a/grafana-kowalski-app/apps/49/errors?frameIndex=0&from=now-7d&to=now&timezone=utc)
2. Refactor the series sorting code to separate concerns into independent functions, each dedicated to a sorting type (alphabetical, outliers, stdDev)

### 📖 Summary of the changes

See diff tab for specific comments

### 🧪 How to test?

- All the automated tests should pass

PR #563, opened 2025-07-22
fix(MetricScene): Small outlier sorting fixes + better E2E tests

grafakus:
### ✨ Description

**Related issue(s):** is a follow-up of https://github.com/grafana/metrics-drilldown/pull/552

This PR makes sure that the sort by outlying series of the `MetricScene` works properly.

### 📖 Summary of the changes

### 🧪 How to test?

- The build should pass with the updated E2E tests.

### DIFF
--- a/e2e/fixtures/components/QuickSearchInput.ts
+++ b/e2e/fixtures/components/QuickSearchInput.ts
@@ -13,7 +13,7 @@ export class QuickSearchInput {
 
   async enterText(searchText: string) {
     await this.get().fill(searchText);
-    await this.page.waitForTimeout(250); // see SceneQuickFilter.DEBOUNCE_DELAY
+    await this.page.waitForTimeout(250);
   }
 
   clear() {


src/services/sorting.ts
```diff
@@ -84,9 +84,9 @@ const getOutliers = (series: DataFrame[]): OutlierOutput => {
 
   // Get number fields: these are our series.
   const joinedSeries = joined.fields.filter((f) => f.type === FieldType.number);
-  const points = joinedSeries.map((series) => new Float64Array(series.values.filter(Boolean)));
```
grafakus, line 87: Filtering out NaN values was a mistake. Indeed, it can produce runtime errors because all Float64Array must have the same length for the detection to work. If they end up containing a different number of values after filtering, an error occurs.

src/services/sorting.ts
```diff
@@ -84,9 +84,9 @@ const getOutliers = (series: DataFrame[]): OutlierOutput => {
...
+  const points = joinedSeries.map((series) => new Float64Array(series.values));
 
-  return OutlierDetector.dbscan({ sensitivity: 0.9 }).preprocess(points).detect();
+  return OutlierDetector.dbscan({ sensitivity: 0.9 }).detect(points);
```
grafakus, line 89: No need for preprocessing. Also the main `sortSeries` is memoized.

src/services/sorting.ts
NWRichmond: Nice catch

NWRichmond (approved):
Very nice! I appreciated the insights in the comments.

From the screenshots, it looks like the sorting is easier to reason about now:

https://github.com/user-attachments/assets/20171f78-c97d-47ba-945c-566903f84c3d

PR #565, opened 2025-07-22
test(EndToEnd): Support multiple Grafana versions for screenshot testing

grafakus:
### ✨ Description

**Related issue(s):** `-`

Prior to this PR, E2E tests could only produce screenshots that would always be matched in CI against **all the versions of Grafana supported by the app**. To make all the E2E tests pass, we chose a [maxDiffPixelRatio option](https://playwright.dev/docs/api/class-snapshotassertions#snapshot-assertions-to-match-snapshot-2-option-max-diff-pixel-ratio) that was [large enough to work across all Grafana versions, and small enough to capture any substantial visual regression](https://github.com/grafana/metrics-drilldown/blob/main/e2e/config/playwright.config.common.ts#L49). 

This is a limitation that this PR aims at removing, by giving us the tools to **generate screenshots for each Grafana version separately**, so that, UI components that vary between Grafana versions can be tested in a more robust way.

To do so, it introduces:
1. a new Playwright fixture `expectScreenshotInCurrentGrafanaVersion()` and
2. a new script `e2e-gen-screenshots.sh` that can be launched locally to generate all the screenshots.

### 📖 Summary of the changes

The major changes are:
- in the `e2e/config` folder, where the Grafana versions supported by the app are defined,
- in the `e2e/fixtures/index.ts` file, where the new `expectScreenshotInCurrentGrafanaVersion()` fixture is defined,
- in the `e2e/tests` folder, where some tests use `expectScreenshotInCurrentGrafanaVersion()` and where new screenshots have been produced,
- in the `scripts` folder, where  the `e2e-gen-screenshots.sh` script has been added

Thanks to these changes, the `maxDiffPixelRatio` option has been divided by 2: from 0.01 to 0.005.

Also, the "Screenshots testing" documentation has been expanded to provide all the information needed to ease the development of new tests.

### 🧪 How to test?

- The build should pass

I highly recommend:
- to checkout this branch, read the docs and test the workflow described by writing a couple of tests :)
- and to check each new screenshot in this PR, as they will serve as a basis for all future builds.


e2e/tests/browser-history.spec.ts
grafakus, line 1: IMO, the confidence that the back button works and will continue fworking in the future is very high now. So we can remove this test.

e2e/tests/metric-scene-view.spec.ts
```diff
@@ -16,11 +16,14 @@ test.describe('Metrics Scene view', () => {
...
   test.describe('Breakdown tab', () => {
-    test('All labels', async ({ metricSceneView }) => {
+    test('All labels', async ({ metricSceneView, expectScreenshotInCurrentGrafanaVersion }) => {
       await metricSceneView.assertDefaultBreadownListControls();
       await metricSceneView.assertPanelsList();
 
-      await expect(metricSceneView.getPanelsList()).toHaveScreenshot('metric-scene-breakdown-all-panels-list.png');
+      await expectScreenshotInCurrentGrafanaVersion(
+        metricSceneView.getPanelsList(),
+        'metric-scene-breakdown-all-panels-list.png'
```
grafakus, line 25: This test was failing with the native `expect().toHaveScreenshot()`. I decided to keep a lower max diff pixel ratio and generate one screenshot per version.

scripts/e2e-del-screenshots.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
grafakus, line 1: Convenience, it'll be useful when a version is not supported anymore.

NWRichmond (approved):
Very nicely done! I appreciate the helpful docs updates.

There are some nice TODOs here, and I'm not too concerned about the current use of timeouts. It's a small price to pay.

I'm curious to see if we have ongoing issues with bleeding-edge `grafana-dev` versions. I'm also curious about how the version upkeep might affect us before leveraging programmatic version numbers.

Overall, I think this is amazing, and is the first step in a direction that I hope will inspire other Grafana apps' e2e testing strategies.

PR #574, opened 2025-07-25
fix(MetricsReducer): Fix zero count when coming back from the MetricScene

grafakus:
### ✨ Description

**Related issue(s):** relates to https://github.com/grafana/metrics-drilldown/pull/550

When selecting a metric in the `MetricsReducer` and coming back (whether by using the browser's back button or clicking on the "Select new metric" button of the `MetricScene`), the counts of metrics in the `QuickSearch` is 0. It seems to happen in non-local environments. 

The cause is similar to the one described in the linked PR.

Here's a short video to see the issue:

https://github.com/user-attachments/assets/9186bc36-c595-4e77-9809-73be1fef9880

### 📖 Summary of the changes

Similar solution to https://github.com/grafana/metrics-drilldown/pull/550.

### 🧪 How to test?

If possible, deploy this PR in a dev instance to ensure it's fixed. Besides, the build should pass.


src/RelatedMetricsScene/RelatedListControls.tsx
```diff
@@ -42,7 +43,7 @@ export class RelatedListControls extends EmbeddedScene {
             body: new QuickSearch({
               urlSearchParamName: 'gmd-relatedSearchText',
               targetName: 'related metric',
-              countsProvider: new MetricVariableCountsProvider(),
+              countsProvider: new MetricVariableCountsProvider() as unknown as CountsProvider,
```
grafakus, line 46: Required to be able to define custom methods on the `MetricVariableCountsProvider` class. I couldn't find a better way to handle this in TypeScript.

PR #576, opened 2025-07-25
perf(*): Prevent duplicated "metrics list" queries when landing on the page

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR reduces the number of queries to fetch the list of all metrics **from 6 to 3**:

| Before | After |
|  ---   |  ---  |
| Screenshot 2025-07-25 at 13 16 50 |  Screenshot 2025-07-25 at 13 18 03 |

Prior to this PR, when landing on the page:
- `DataTrail` would render its component,
- during render, [histograms would be initialized](https://github.com/grafana/metrics-drilldown/blob/main/src/DataTrail.tsx#L310) by calling [MetricDatasourceHelper.initializeHistograms()](https://github.com/grafana/metrics-drilldown/blob/main/src/helpers/MetricDatasourceHelper.ts#L114),
- `MetricDatasourceHelper.initializeHistograms()`would make [2 requests](https://github.com/grafana/metrics-drilldown/blob/main/src/helpers/MetricDatasourceHelper.ts#L117-L118) to get the data needed,
- several render cycles of the `DataTrail` component would occurs, resulting in duplicated queries.

Furthermore, [FilteredMetricsVariable](https://github.com/grafana/metrics-drilldown/blob/main/src/WingmanDataTrail/MetricsVariables/FilteredMetricsVariable.ts), which extended from `MetricsVariable`, would also make its own query when mounted.

### 📖 Summary of the changes

1. During the render cycle of the `DataTrail`, histograms are now initialized within `useEffect`,
2. `FilteredMetricsVariable` does now extend from `CustomVariable` and listens to the change of state of  `MetricsVariable`

### 🧪 How to test?

- See the browser's "Network" tab before and after checking out this PR
- The build should pass

### DIFF
--- a/src/DataTrail.tsx
+++ b/src/DataTrail.tsx
@@ -186,8 +186,8 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
     if (!this.state.histogramsLoaded) {
       try {
         await this.datasourceHelper.initializeHistograms();
-      } catch {
-        displayWarning(['Error while initializing histograms!']);
+      } catch (e) {
+        displayWarning(['Error while initializing histograms!', (e as Error).toString()]);
       }
 
       this.setState({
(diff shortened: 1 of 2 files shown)


src/DataTrail.tsx
```diff
@@ -306,8 +306,11 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
     const chromeHeaderHeight = useChromeHeaderHeight() ?? 0;
     const headerHeight = embedded ? 0 : chromeHeaderHeight;
     const styles = useStyles2(getStyles, headerHeight, model);
+
     // need to initialize this here and not on activate because it requires the data source helper to be fully initialized first
-    model.initializeHistograms();
+    useEffect(() => {
+      model.initializeHistograms();
+    }, [model]);
```
yangkb09, line 313: nice!

bohandley (commented):
LGTM! Thank you Marc!

PR #595, opened 2025-07-30
fix: add support for panels with very small values in data

### LINKED ISSUE #354: [Bug bash] - Update the queries to prevent NaN values from showing
Image

NWRichmond:
### ✨ Description

**Related issue(s):** This PR fixes #354.

In Prometheus, things can get weird when dealing with extremely small values. Sometimes, a query like `avg(some_metric)` can produce all `NaN` values, leading to empty panels, which can be confusing because individual series have valid values. For example:

https://github.com/user-attachments/assets/62c21027-2f2d-4458-a709-a0e86c179cfc

So what's happening here? When averaging extremely small values (e.g., `9e-129`) with "normal-sized" numbers, intermediate arithmetic operations can [underflow](https://en.wikipedia.org/wiki/Arithmetic_underflow) below the representable float64 range, causing the aggregation result to become `NaN`. This occurs because Prometheus uses standard IEEE 754 floating-point arithmetic, which loses precision when operating on values with vastly different magnitudes.

So what can we do about it? First, we can detect that this is happening using a [Scenes behavior](https://grafana.com/developers/scenes/advanced-behaviors) that inspects the data returned by the Prometheus data source. Next, instead of `avg(some_metric)`, we can re-run the query with the expression `avg(some_metric and (some_metric > -Inf))`.

`avg(some_metric and (some_metric > -Inf))` helps by filtering out non-finite values before aggregation. The `and` operator performs label-based filtering, excluding any series where `some_metric` contains extremely small values that lead to underflow. By removing these problematic values upfront, the averaging calculation operates only on finite numbers, preventing the arithmetic operations from propagating `NaN` results that would otherwise contaminate the entire aggregation.

### 📖 Summary of the changes

This PR introduces a new `extremeValueFilterBehavior` that deals with the problem described above. We implement this in both the Metrics Reducer and the Metric scene.

### 👀 Demo

Compare this to the video above:

https://github.com/user-attachments/assets/1babfa7a-8bde-49f9-aa1d-3b05762416a4

### 🧪 How to test?

I found this is most easily tested with the [**prometheus-otel** datasource](https://start.1password.com/open/i?a=WVKUHDMXXVCZ7M6OLE3Y52MX6I&v=jb3rz727znuwwvbl3v2mmdnigm&i=bo3f5kmilazdebxisc4kxzynum&h=grafana.1password.com), which has a few metrics that begin with `appo11y:duration:p95` that demonstrate this issue.


src/autoQuery/behaviors/ExtremeValueFilterBehavior.ts
```diff
@@ -0,0 +1,249 @@
...
+ * Implementing this as a behavior allows us to apply the extreme values filtering
+ * to only the queries that are affected by the extreme values, rather than all queries in the scene.
+ * That way, we can keep queries simpler by default, and only apply the extreme values filtering
+ * when it's necessary.
+ */
+export function extremeValueFilterBehavior(sceneObject: SceneObject): CancelActivationHandler | void {
+  const [queryRunner] = sceneGraph.findDescendents(sceneObject, SceneQueryRunner);
+
+  if (!queryRunner) {
+    logger.warn('ExtremeValueFilterBehavior: No query runner found in parent scene');
```
grafakus, line 29:
```suggestion
    logger.warn('ExtremeValueFilterBehavior: No query runner found in descendents of the current scene object');
```

NWRichmond: oops, a leftover from an earlier state...fixed in https://github.com/grafana/metrics-drilldown/pull/595/commits/7b70fc3bc1485844f28cc7a9b91d40b414f7d889 :)

NWRichmond:
Hi @grafakus! I've made some updates:
- The check for `NaN` now leverages the `"NaN"` key in the `series[0].fields[1].entities`, to reduce complexity of the check
- There's a new event so we can track how often the `extremeValueFilterBehavior` runs
- There's now a `statusMessage` on panels when the `extremeValueFilterBehavior` "heals" the query:

https://github.com/user-attachments/assets/5ef1987e-cf3a-40c5-9cc7-cd89a9bf46e1

demo panel statusMessage metric scene

Finally, I created https://github.com/grafana/metrics-drilldown/issues/597 to capture follow-up work on this, to allow the user more control.

grafakus (commented):
Thank you for the PR Nick! Unfortunately, it looks like something breaks the app :( As I was testing with this branch in local, I had a couple of these errors:
image

Sometimes they appear after scrolling a bit in the `MetricsReducer`.

Also, maybe we should consider to decrease the usage of the `Logger`? For instance, if we already sent a Rudderstack event, why do we also need to log an info message? 

Can we also motify the user if `extremeValueFilterBehavior` cannot complete its job?

It brings an important practical point: why do we need these log messages? Do we plan to review all the existing and upcoming warning messages? Considering that we may not be able to act on them (e.g. "ExtremeValueFilterBehavior: No queries found in query runner"), my recommendation would be to:
- Display a warning message on the panel when `extremeValueFilterBehavior` was not able to perform its job
- Log a warning in this case with the cause
- Remove all other `Logger` calls

src/autoQuery/behaviors/ExtremeValueFilterBehavior.tsx
```diff
@@ -0,0 +1,300 @@
...
+ * @remarks
+ * Implementing this as a behavior allows us to apply the extreme values filtering
+ * to only the queries that are affected by the extreme values, rather than all queries in the scene.
+ * That way, we can keep queries simpler by default, and only apply the extreme values filtering
+ * when it's necessary.
+ */
+export function extremeValueFilterBehavior(panel: VizPanel): CancelActivationHandler | void {
+  const [queryRunner] = sceneGraph.findDescendents(panel, SceneQueryRunner);
+
+  // Prevent double-unsubscribing, by tracking if the subscription has been unsubscribed already.
```
grafakus, line 32: Why having to keep track of it instead of unsubscribing when the query runner is replaced by its clone?

grafakus (approved):
👏🏾 Great job Nick!
Let's merge it today so it can be included in our weekly release.

Once we have collected enough data from our users, let's follow-up on the UX aspects in https://github.com/grafana/metrics-drilldown/issues/597

PR #612, opened 2025-08-06
fix: add types for legacy language provider methods

### LINKED ISSUE #575: Deprecation of prometheus language provider methods
We recently deprecated some methods in languageProvider in order to bring a unified experience. 
See PR: https://github.com/grafana/grafana/pull/105818

In prometheus package all usages of old API is removed. And soon deprecated APIs will be deleted. 

It's better to start using new API if possible https://github.com/grafana/grafana/blob/065552338d4372a799fd1b30cccfe0bcccf54ef0/packages/grafana-prometheus/src/language_provider.ts#L518-L590

NWRichmond:
### ✨ Description

**Related issue(s):** This PR fixes #575.

## Context

`@grafana/prometheus`' `languageProvider` methods have evolved a lot from `@grafana/prometheus` version 11.6.0 to present (currently 12.1.x). Because the Metrics Drilldown app supports Grafana versions back to 11.6.x ([until 2025-06-25](https://grafana.com/docs/grafana/latest/upgrade-guide/when-to-upgrade/#what-to-know-about-version-support)), we have to jump through some hoops to accommodate the multiple flavors of language provider methods available `@grafana/prometheus` in versions 11.6.x, 12.0.0, and 12.1.0+. These techniques are documented in [`MetricDatasourceHelper`](https://github.com/grafana/metrics-drilldown/blob/main/src/helpers/MetricDatasourceHelper.ts)'s code comments.

`@grafana/prometheus` will soon undergo another change to prepare for a cleaner, simpler future with the language provider. In https://github.com/grafana/grafana/pull/108352, the language provider methods that were deprecated in https://github.com/grafana/grafana/pull/105818 will be removed. Metrics Drilldown is already prepared for this at runtime – we use the newer language provider methods, also introduced in https://github.com/grafana/grafana/pull/105818, when they exist on the runtime datasource's language provider.

That said, we should expect that upgrading Metrics Drilldown's `@grafana/prometheus` version after https://github.com/grafana/grafana/pull/108352 is merged will cause compile-time issues, as TypeScript will not be aware of the existence of the deprecated language provider methods that we still use at runtime for users of Grafana 11.6.x - 12.0.x.

This PR helps avoid those compile-time issues before we encounter them, by augmenting the `@grafana/prometheus` type definitions for the language provider.

### 📖 Summary of the changes

1. Adds Language Provider types from `@grafana/prometheus` versions `11.6.0` and `12.0.0` to help the `MetricDatasourceHelper` precisely know the methods (and their signatures) that are available in the version of the Prometheus data source that's made available by `@grafana/runtime`.
2. Refactors `MetricDatasourceHelper` to simplify the language provider version checks. Checks with `dsUsesTimeRangeInLegacyLanguageProviderMethods` are no longer needed!

### 🧪 How to test?

All CI should pass.

### DIFF
--- a/src/helpers/MetricDatasourceHelper.ts
+++ b/src/helpers/MetricDatasourceHelper.ts
@@ -237,15 +237,14 @@ export class MetricDatasourceHelper {
    */
   public static fetchLabels(params: FetchLabelsOptions): Promise<string[]> {
     const { timeRange, matcher } = params;
-    const ds = params.ds as PrometheusRuntimeDatasource;
+    const ds = params.ds;
 
     if (languageProviderVersionIs['12.1.0-plus'](ds)) {
       return ds.languageProvider.queryLabelKeys(timeRange, matcher);
     } else if (languageProviderVersionIs['12.0.0'](ds)) {
-      // eslint-disable-next-line sonarjs/deprecation
       return ds.languageProvider.fetchLabelsWithMatch(timeRange, matcher).then((labels) => Object.keys(labels));
     } else if (languageProviderVersionIs['11.6.x'](ds)) {
-      return ds.languageProvider.fetchLabelsWithMatch(matcher).then((labels) => Object.keys(labels)); // eslint-disable-line sonarjs/deprecation
+      return ds.languageProvider.fetchLabelsWithMatch(matcher).then((labels) => Object.keys(labels));
     }
 
     throw new Error('Unsupported language provider version');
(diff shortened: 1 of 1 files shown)


src/helpers/MetricDatasourceHelper.ts
```diff
@@ -251,19 +236,19 @@ export class MetricDatasourceHelper {
...
+    if (languageProviderVersionIs['12.1.0-plus'](ds)) {
       return ds.languageProvider.queryLabelKeys(timeRange, matcher);
-    }
-
-    if (MetricDatasourceHelper.dsUsesTimeRangeInLegacyLanguageProviderMethods(ds)) {
+    } else if (languageProviderVersionIs['12.0.0'](ds)) {
       // eslint-disable-next-line sonarjs/deprecation
       return ds.languageProvider.fetchLabelsWithMatch(timeRange, matcher).then((labels) => Object.keys(labels));
+    } else if (languageProviderVersionIs['11.6.x'](ds)) {
+      return ds.languageProvider.fetchLabelsWithMatch(matcher).then((labels) => Object.keys(labels)); // eslint-disable-line sonarjs/deprecation
```
grafakus, line 248:
```suggestion
    } else if (languageProviderVersionIs['12.0.0'](ds)) {
      return ds.languageProvider.fetchLabelsWithMatch(timeRange, matcher).then((labels) => Object.keys(labels));
    } else if (languageProviderVersionIs['11.6.x'](ds)) {
      return ds.languageProvider.fetchLabelsWithMatch(matcher).then((labels) => Object.keys(labels)); // eslint-disable-line sonarjs/deprecation
```

src/helpers/MetricDatasourceHelper.ts
```diff
@@ -251,19 +236,19 @@ export class MetricDatasourceHelper {
    * ```
    */
   public static fetchLabels(params: FetchLabelsOptions): Promise<string[]> {
-    const { ds, timeRange, matcher } = params;
+    const { timeRange, matcher } = params;
+    const ds = params.ds as PrometheusRuntimeDatasource;
```
grafakus, line 240:
```suggestion
    const ds = params.ds;
```

The type assertion is not necessary if we modify the `FetchLabelsOptions` below. And we can do the same for `FetchLabelValuesOptions`.

src/helpers/MetricDatasourceHelper.ts
```diff
@@ -296,23 +281,32 @@ export class MetricDatasourceHelper {
...
+
+    if (languageProviderVersionIs['11.6.x'](ds)) {
+      // If a matcher isn't provided, use the simpler `fetchLabelValues` method.
+      const fetchLabelValuesWithOptionalMatcher = matcher
+        ? ds.languageProvider.fetchSeriesValuesWithMatch // eslint-disable-line sonarjs/deprecation
+        : ds.languageProvider.fetchLabelValues; // eslint-disable-line sonarjs/deprecation
 
-    if (MetricDatasourceHelper.dsUsesTimeRangeInLegacyLanguageProviderMethods(ds)) {
-      return fetchLabelValuesWithOptionalMatcher(timeRange, labelName, matcher);
+      return fetchLabelValuesWithOptionalMatcher(labelName, matcher); // eslint-disable-line sonarjs/deprecation
```
grafakus, line 306:
```suggestion
      return fetchLabelValuesWithOptionalMatcher(labelName, matcher); 
```

grafakus (dismissed):
Great job Nick!

I left only a suggestion to remove some type assertions that are not needed anymore and to remove all the `eslint-disable` comments.

bohandley (dismissed):
LGTM! 🚀 Thank you for working on this, keeping our Prometheus dependencies well kept!

NWRichmond:
@grafakus thanks for the suggestions! I'm not sure why my local setup was preventing those comments from being removed without errors last week, but it's working fine now without them.

Issue #450, opened 2025-05-15
[spike] Determine which functions and parameters we provide to the user per panel type


grafakus:
**Acceptance criteria:**
- For each panel type, come up with a list of relevant functions to provide to the user
- For the functions that can be parametrized (e.g. percentiles) come up with a list of parameters that can also help our users (e.g. P99, P50, P20, ...)
- Consider also if it makes sense to add some global settings to the app
- Consider adding `or vector(0)` to the `avg` queries or “connect zero values” (+ add  'missing data' tooltip when hovering)? ([see Slack thread](https://raintank-corp.slack.com/archives/C0630AYC4SY/p1741903178447189))
- Consider this [use case shared in Slack](https://raintank-corp.slack.com/archives/C075BDBTX96/p1753385029151679), where a gauge metric with a value measured as a Unix timestamp in seconds is queried (`time() - <metric>`) to get an age

NWRichmond:
> About the Stat panel for statuses, I was thinking that, at a glance, our users could see which metrics are currently up or down, as well as how the status evolves in the background

Cool! I wonder how we can present this...seems like it'd be challenging in the Metrics Reducer, but easier in the Metrics scene with more real estate.

Issue #425, opened 2025-05-13
Complete the implementation of the "Configure Prometheus function" feature


grafakus:
**Acceptance criteria:**
1. The "Configure" button has been added to each panel:
<img width="350" alt="Image" src="https://github.com/user-attachments/assets/dac00a39-4023-4da5-ad21-0449386af388" />

2. Once clicked, it opens a drawer with a list of viz panels that allow the user to customize the function used:
<img width="350" alt="Image" src="https://github.com/user-attachments/assets/ab4c091e-729b-4c8f-a0d0-f8d6bf07202a" />

3. After applying a new function the viz panel is updated
4. **[TO BE DEFINED]** The new configuration is persisted, where? In local storage?
5. When navigating around the app, the configuration is automatically applied (in the metrics reducer, the metric scene, the related metrics, ...)
6. Some End-To-End tests have been implemented

**Note:**

To simplify the work, we might want to refactor the app first, so that it uses only a single panel component ([MetricVizPanel](https://github.com/grafana/metrics-drilldown/blob/main/src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx)?)

PR #594, opened 2025-07-29
refactor(*): Use single VizPanel across the app

grafakus:
### ✨ Description

**Related issue(s):** resolves https://github.com/grafana/metrics-drilldown/issues/448, blocks https://github.com/grafana/metrics-drilldown/pull/613

This PR is a refactor to unify how we build metric viz panels across the app. It has been presented in details to the team (see the [meeting notes and recording](https://docs.google.com/document/d/1e0LjG9XR-mNNv3dtcZq-c7fRUxEl2PqxOPQGM-72ITc/edit?tab=t.kh04aaacrrc3))

### Context

Where and how do we currently build panels and their queries?

**A. Wingman (6 months ago)**

1. Panels: [src/WingmanDataTrail/MetricVizPanel/panels](src/WingmanDataTrail/MetricVizPanel/panels)
2. Queries: [src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx](src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx) (`determineQueryProperties` and [src/autoQuery/buildPrometheusQuery.ts](src/autoQuery/buildPrometheusQuery.ts))

Queries are built in a "modern way" with the [grafana/tsqtsq library](https://github.com/grafana/tsqtsq)

**B. autoQuery (legacy)**

1. Panels: [src/autoQuery/graphBuilders.ts](src/autoQuery/graphBuilders.ts)
2. Queries: [src/autoQuery/getAutoQueriesForMetric.ts](src/autoQuery/getAutoQueriesForMetric.ts)

Actually `getAutoQueriesForMetric` does it all:

- queries, in a "non-modern way" (strings manipulation)
- panels and _all their variants_

On top of it, [src/autoQuery/components/AutoVizPanel.tsx](src/autoQuery/components/AutoVizPanel.tsx) is a Scene object that can render any panel type and add a selector in the header actions to switch variant (only used for heatmap/percentiles)

### Combining the best of both worlds

This PR aims at using all the good ideas existing in the code, namely:

- The legacy "Auto" mechanism: given a metric, it produces a panel and its queries
- Query building with the `grafana/tsqtsq` library
- Separation of concerns (layers): panel component → panel builder → queries builder for
  - Maintainability
  - Testeability
  - Debuggeability
  - Performance optimizability
  - Modularity
  - Extensibility

[src/GmdVizPanel/GmdVizPanel.tsx](src/GmdVizPanel/GmdVizPanel.tsx) (the name is temporary)

[...]

PR #613, opened 2025-08-07
feat(MetricScene): Enable users to configure the Prometheus function and panel type for any given metric

grafakus:
### ✨ Description

**Related issue(s):** solves https://github.com/grafana/metrics-drilldown/issues/425 and relates to  https://github.com/grafana/metrics-drilldown/issues/450

| Metric type | Screenshot |
|  ---   |  ---  |
| **Gauge** | image |
| **Counter** | image |
| **Age** | image |
| **Histogram** | image |
| **Status up/down** | image |

### 📖 Summary of the changes

**Changes in `GmdVizPanel`**

- The constructor of `GmdVizPanel` now receives its options in a more structured way via :`panelOptions` (to customize the viz panel) and `queryOptions` (to customize the query).
- `GmdVizPanel` is now responsible for retrieving any user preferences and applying them automatically.

**The new `ConfigurePanelForm` component**

- `ConfigurePanelForm` is responsible for displaying the config form with a time picker and a "Restore default config" on top, a gallery of config presets in the body and the "Apply" and "Cancel" buttons in the footer. Once applied, the selected preset is saved as a user preference.
- Within the form, each preset is rendered as a decorated `GmdVizPanel`, which is taken care of by the new decorator component `WithConfigPanelOptions`.
- `WithConfigPanelOptions` adds a radio button as well as controls for query parameters, if any (like percentiles).
- `ConfigurePanelForm` is rendered in a drawer by `MetricScene`, which listens to some events in order to open/close the drawer (`EventConfigurePanel`/`EventCancelConfigurePanel`) and to apply the preset chosen by the user (`EventApplyPanelConfig`).

**Presets per metric type**

A set of presets have been created for configuring the panel options (like its title or description) and the query options (like the Prometheus function to use). They are grouped per metric type:`ages`, `histograms`, `status-updown`, `timeseries`.

- All the code related to the presets is located in the folder `src/GmdVizPanel/config/presets`. There's one file per metric type.
- All the custom PromQL functions supported by the app are available in `src/GmdVizPanel/config/promql-functions.ts`, which contains a map between function names (that we store in the user preferences) and its equivalent in Javascript (which we use to build the query).
- The code to build all the different queries now uses this map, which makes possible to apply any preset.

**User actions tracking**

- The main user actions are tracked:`configure_panel_opened`, `default_panel_config_restored `, `panel_config_applied `, with the metric type (and the config id when it makes sense).

### 🧪 How to test?

Locally, after checking this branch, to get a feeling of the feature as a whole. Try to find a metric of each type (see `src/GmdVizPanel/config/presets/getConfigPresetsForMetric.ts` to understand how they are identified by the app).

Also, as usual, the build should pass with the new unit and E2E tests with updated screenshots.


src/GmdVizPanel/GmdVizPanel.tsx
```diff
@@ -1,289 +1,270 @@
...
-    if (isNativeHistogram || isHistogramMetric(metric)) {
-      return PANEL_TYPE.HEATMAP;
+    if (histogramType === 'classic' || histogramType === 'native') {
+      return 'heatmap';
     }
 
-    return PANEL_TYPE.TIMESERIES;
-  }
-
-  public static getPanelHeightInPixels(h: PANEL_HEIGHT): number {
```
grafakus, line 163: enums FTW

grafakus:
Thank you for taking the time to test it. It's a rather large PR so I really appreciate your time and feedback! 🤗

> Users might find it unintuitive that they need to scroll to save their function config:
> I wonder if it could make sense to move this up next to Restore default config?

I initially chose to place them in a predictable place, the reasoning being that 99% of the forms out there have them at the bottom of the form. What do you think about by making them sticky in https://github.com/grafana/metrics-drilldown/pull/613/commits/e7566f4ff2c9418e7f43a41f9cf92cee32dd4bb0. 

image

> I found myself wishing this said something like, "Are you sure you want to restore the default configuration for the `${metric}` metric?".

Sure, it's been done in https://github.com/grafana/metrics-drilldown/pull/613/commits/6e43f3746dda87b327cc2c9e8c06fc8b122f8106:

image

> If a user configures a large number of panels and wants to restore their defaults, this can be challenging (time consuming at best, "oh nooo which panels did I configure?!?" at worst). I could see a **Restore all defaults** button (or similar) helping with this. Our sidebar's **Settings** tab is empty, but this could be a helpful first item in that tab.

That's a great idea! Let's measure the usage of this first version and, based on the qualitative and quantitative feedback received from our users, we can plan the next iterations.

src/GmdVizPanel/types/timeseries/buildTimeseriesPanel.ts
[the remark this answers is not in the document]
grafakus: The data shows up indeed but it's very confusing to see two `i` icons in different colours. How do you suggest we solve this confusion?

src/WingmanDataTrail/MetricVizPanel/MetricVizPanel.tsx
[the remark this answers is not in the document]
grafakus: I think it's just to satisfy TS. This file will be deleted as part of the cleanup post https://github.com/grafana/metrics-drilldown/pull/594, so we shouldn't worry too much.

src/GmdVizPanel/components/ConfigurePanelAction.tsx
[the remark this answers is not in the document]
bohandley: This is a good point! We can take this to [Jess Matz for the meeting on the calls to action changes](https://raintank-corp.slack.com/archives/C0630AYC4SY/p1754423741910239?thread_ts=1754000074.379019&cid=C0630AYC4SY). The metric function configuration is definitely a panel specific action, but maybe there is another type of icon we can replace the gear with in the meantime.

bohandley (approved):
This is incredible! So many great elements to this work. 🎉 Love it!!!

This is my understanding of it:

- User clicks "Configure" on a metric panel
- System determines appropriate presets based on metric type (counter, histogram, age, etc.)
- Configuration drawer opens showing preset options with live previews
- User selects desired function/panel type (e.g., "Sum" vs "Average" for timeseries)
- User clicks "Apply" - configuration is saved to user preferences
- Panel updates to use the selected configuration
- Future loads of that metric will use the saved preference

The only feature I can think of to add to this would be something @NWRichmond already mentioned, a "reset all configurations." But then this is generally adjusting the user preferences, so maybe we could have the "reset all" live in a future admin dropdown that lists the user preferences?

```
grafana-metricsdrilldown-app.bookmarks	
grafana-metricsdrilldown-app.datasource	
grafana-metricsdrilldown-app.grafana.trails.recent
grafana-metricsdrilldown-app.metric-prefs	
grafana-metricsdrilldown-app.recent-metrics
```

File structure looks nice too, code organization makes sense! This is a huge PR but I'm happy to have it all together to see it all work nicely 💯 

The main organzation I see is

1. Configuration System in `GmdVizPanel/config/` (love the presets)
2. Configuration UI in `GmdVizPanel/components/ConfigurePanelForm/`
3. State Management & Persistence in `UserPreferences/`
4. Panel Integration in the Metric Scene and Metric Reducer

This is great work, I think that users will love it and we'll get all sorts of feedback about particular functions and additions in the future.

src/GmdVizPanel/components/GmdVizPanelVariantSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+  metric: string;
+  options: Array<{
+    label: string;
+    value: PanelType;
+  }>;
+  currentPanelType?: PanelType;
+}
+
+// currently used only for histogram metrics
+export class GmdVizPanelVariantSelector extends SceneObjectBase<GmdVizPanelVariantSelectorState> {
```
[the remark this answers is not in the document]
grafakus: > I would argue that this could be remove 🗑️ as you can select the percentiles or the heatmap the function config

I have the same opinion. I've added some user tracking in https://github.com/grafana/metrics-drilldown/pull/613/commits/9bbb91f0e5bd9becb731cb3400278d4369960803 so we can based our decision on user data.

grafakus:
> This looks amazing overall! The only blocker so far is that the [`extremeValueFilterBehavior`](https://github.com/grafana/metrics-drilldown/pull/595/files#diff-080b59bfbedbb3bcb6ebb1dd2abdd086e6a8b2c8e4cefeab73c87dcd667eeb80) is still missing from panels, which creates a regression of #354. Let's add the behavior back and address the tooltip UX concerns as a follow-up.

@NWRichmond Sorry, my bad 🤦🏾‍♂️

It's been added in https://github.com/grafana/metrics-drilldown/pull/613/commits/01652d9b69fe479687e143ac2541e3f3e948d6c0 for all timeseries. Is it ok to leave the percentiles queries without?

src/DataTrail.tsx
```diff
@@ -27,14 +27,23 @@ import {
 import { useStyles2 } from '@grafana/ui';
 import React, { useEffect } from 'react';
 
+import { ConfigurePanelAction } from 'GmdVizPanel/components/ConfigurePanelAction';
+import { ConfigurePanelForm } from 'GmdVizPanel/components/ConfigurePanelForm/ConfigurePanelForm';
+import { EventApplyPanelConfig } from 'GmdVizPanel/components/ConfigurePanelForm/EventApplyPanelConfig';
+import { EventCancelConfigurePanel } from 'GmdVizPanel/components/ConfigurePanelForm/EventCancelConfigurePanel';
+import { EventConfigurePanel } from 'GmdVizPanel/components/EventConfigurePanel';
+import { GmdVizPanel } from 'GmdVizPanel/GmdVizPanel';
+import { getMetricType } from 'GmdVizPanel/matchers/getMetricType';
```
shelldandy, line 36: nit: i think the `GmdVizPanel/components` could have an index file and that way we can just import from a single place instead of these many lines right here.

src/GmdVizPanel/types/available-panel-types.ts
```diff
@@ -0,0 +1,11 @@
+export const AVAILABLE_PANEL_TYPES = [
+  'heatmap',
+  'percentiles',
+  'stat',
+  'statushistory',
+  'timeseries',
+  // TODO for info metrics, see https://github.com/grafana/metrics-drilldown/issues/450
+  // 'table',
+] as const;
```
shelldandy, line 9: Are these types not available in `@grafana/data` or `@grafana/schema`

src/GmdVizPanel/config/promql-functions.ts
```diff
@@ -0,0 +1,68 @@
+import { promql } from 'tsqtsq';
+
+export const PROMETHEUS_FUNCTIONS = [
```
shelldandy, line 3: nit: most of these are actually aggregators so it could be confusing

https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators

shelldandy (dismissed):
i had a couple of very minor comments and nits but overall this is a fantastic addition to GMD with great UX thanks Marc!

bohandley (dismissed):
LGTM! I love the structure and the organization. Thank you Marc!

NWRichmond (approved):
Very nicely done! 👏🏼 👏🏼 👏🏼 

This is an ambitious feature, and I expect that users will be delighted by it. It's complex, but the complexity is manageable and there are solid conventions for extending this in the future.

We've got a good list of follow-up items, but nothing worth waiting over. This PR provides a solid foundation.

It looks like something went wrong in https://github.com/grafana/metrics-drilldown/commit/f8136519dd4dcff205229f7c8a447f3372878102, but once that's fixed, I think this is ready to roll :)

yangkb09 (dismissed):
this is a really exciting feature, marc! thank you for the walkthrough too!

src/DataTrail.tsx
grafakus: I'm always a bit reluctant because of https://vercel.com/blog/how-we-optimized-package-imports-in-next-js
Wdyt?

src/GmdVizPanel/types/available-panel-types.ts
grafakus: I'm not sure about the advantage, knowing that we want to able to define any panel type that makes sense for the app. In the future, we may want to define a "precentiles-with-cool-stuff" panel.

src/GmdVizPanel/config/promql-functions.ts
grafakus: Aren't they also functions?

e2e/tests/metric-scene-view.spec.ts-snapshots/metric-scene-related-metrics-all-list-chromium-linux.png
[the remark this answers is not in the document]
grafakus: There are more metrics 🤔

Issue #615, opened 2025-08-07
[bug] Fatal error in the MetricScene after selecting a label


grafakus:
**Environment:** local/ops
**App version:** 1.0.8
**Grafana version:** 11.6.4 to 12.2.0

**Steps to reproduce:**
1. Go to [http://localhost:3001/a/grafana-metricsdrilldown-app/drilldown?from=now-1h&to=now&timezone=utc&nativeHistogramMetric=1&layout=grid&filters-rule=&filters-prefix=grafana&filters-suffix=&var-filters=&var-labelsWingman=%28none%29&search_txt=grafana_database_all_migrations_duration_seconds&var-metrics-reducer-sort-by=default&var-ds=gdev-prometheus&var-other_metric_filters=filters-prefix,prefix%7C%3D%7Cgrafana&metric=grafana_database_all_migrations_duration_seconds&actionView=breakdown&var-groupby=$__all&breakdownLayout=grid
3. Add a (e.g.) `instance=xyz` ad hoc filter](http://localhost:3001/a/grafana-metricsdrilldown-app/drilldown?from=2025-05-26T11:00:00.000Z&to=2025-05-26T12:05:00.000Z&timezone=utc&nativeHistogramMetric=1&layout=grid&filters-rule=&filters-prefix=&filters-suffix=&var-filters=instance%7C%3D%7Cgrafana-gmd:3000&var-labelsWingman=%28none%29&search_txt=&var-metrics-reducer-sort-by=default&var-ds=gdev-prometheus&var-other_metric_filters=&metric=grafana_database_all_migrations_duration_seconds&actionView=breakdown&var-groupby=$__all&breakdownLayout=grid)
2. Select the panel with the title "instance (1)":

<img width="566" height="367" alt="Image" src="https://github.com/user-attachments/assets/33f21a53-c107-4d5a-a083-29a053699c6c" />

**What happens**

A fatal error:
<img width="1072" height="899" alt="Image" src="https://github.com/user-attachments/assets/e8eac089-ba83-444c-9c5d-0cfabe92b236" />

**What should have happened**

The "Select" button should not be rendered:

<img width="567" height="301" alt="Image" src="https://github.com/user-attachments/assets/ad80e733-cb13-449e-b6f3-e490379ba1a3" />

If the user lands via a previously shared link and the label value is already selected, there should not be any fatal error, but an empty state:

[...]

PR #616, opened 2025-08-07
fix(MetricScene): Fix fatal error in an edge case where there's no data

grafakus:
### ✨ Description

**Related issue(s):** fixes https://github.com/grafana/metrics-drilldown/issues/615

### 📖 Summary of the changes

Two small changes:
1. We disable the "Select" button when there's no data
2. We improve the sorting code to prevent a runtime exception when the data frame is empty 

### 🧪 How to test?

- Locally, after pulling this branch. See the linked issue for the reproduction steps.


bohandley (approved):
This was such a good catch! Nicely done @grafakus !!

PR #608, opened 2025-08-06
refactor(MetricLabelValuesList): Introduce LabelValueVizPanel

grafakus:
### ✨ Description

**Related issue(s):** preparation to simplify the review on https://github.com/grafana/metrics-drilldown/pull/594

This PR adds the new `LabelValueVizPanel` component to encapsulate how we build panel in the "Breakdown" tab after selecting a label:
image

This is made to simplification the panel unification in https://github.com/grafana/metrics-drilldown/pull/594

### 📖 Summary of the changes

The code for building the panel has been encapsulated in a new Scene object `LabelValueVizPanel`

### 🧪 How to test?

- The build should pass

PR #617, opened 2025-08-07
fix(Breakdown): Address Scene warnings

grafakus:
### ✨ Description

**Related issue(s):** is caused by https://github.com/grafana/metrics-drilldown/pull/608

After merging the linked PR, some warnings started to appear in the browser console, for instance:

image

### 📖 Summary of the changes

This PR modifies `LabelValueVizPanel` so that it does not store any Scene object in its own state. By doing so, they can be used to build the timeseries panel (and attached only to it in the Scene graph).

### 🧪 How to test?

- No warning messages related to Scene objects with multiple parents should appear in the browser's console


NWRichmond (approved):
Nice catch!

Issue #621, opened 2025-08-11
[bug] AdHoc filters containing values with special chars like the equal sign (=) are not applied properly


grafakus:
Issue [reported in Slack](https://raintank-corp.slack.com/archives/C075BDBTX96/p1754921364385849) by @Logiraptor

**ROOT CAUSE:**

Our [custom expression builder for the AdHocFiltersVariable ](https://github.com/grafana/metrics-drilldown/blob/main/src/DataTrail.tsx#L404-L410) should escape label values [as Scenes does](https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/utils.ts#L77-L79).

**FIX:**

1. **Very quick fix:** escape all equal signs in our expression builder
2. **Quick fix:** the function Scenes uses to escape the label values is not exposed, so we could duplicate the code from Scenes in our repo
3. **Optimal fix:** consider pinging the Scenes owners to see if they are willing ot expose the utility function so we don't have to duplicate the code in our repo

grafakus:
https://github.com/grafana/metrics-drilldown/pull/623 provided a specific solution, https://github.com/grafana/metrics-drilldown/pull/657 provided a generic solution.

Let's verify the generic fix works after releasing v1.0.13

PR #623, opened 2025-08-11
fix(AdHocFilters): Escape equal signs in label values

grafakus:
### ✨ Description

**Related issue(s):** relates to https://github.com/grafana/metrics-drilldown/issues/621

This PR is a very quick fix for https://github.com/grafana/metrics-drilldown/issues/621. It's not the best thing we can do (the alternatives are explained in the related issue) but at least it makes the app usable.

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- The build should pass, with the new unit tests


bohandley (approved):
LGTM!

PR #635, opened 2025-08-18
feat: make related metrics search bar sticky

yangkb09:
### ✨ Description

**Related issue(s):** 

Make the related metrics search bar sticky. This will be better UX since users can filter even after scrolling down the list of related metrics.

Before

https://github.com/user-attachments/assets/b42a1aa0-aab3-4020-9816-f9390d2443fc

After

https://github.com/user-attachments/assets/3ebb6145-450f-4b10-b301-d6be6705eaf6

Fixes https://github.com/grafana/grafana/issues/94453?reload=1?reload=1

### 📖 Summary of the changes

- [ ] Separated the `MetricsActionBar` (breakdown, etc. tabs) from the main metric panel (topView) so that they can be conditionally sticky
- [ ] Deprecated the main metric panel sticky feature, as agreed upon in meetings

### 🧪 How to test?

- Build passes
- Open a metric
- Switch to the "Related metrics" tab
- Scroll down the list of related metrics, confirm that the related metrics search bar is sticky and that you can use it to search


src/MetricGraphScene.tsx
```diff
@@ -160,3 +178,14 @@ function getStyles(theme: GrafanaTheme2, headerHeight: number, trail: DataTrail)
     }),
   };
 }
+
+function updateActionBarHeight() {
```
shelldandy, line 182: i think this function should also be updated to use the ref instead of querySelector.

So something like:

const actionBar = controlsContainer.current;

src/RelatedMetricsScene/RelatedMetricsScene.tsx
```diff
@@ -152,15 +158,20 @@ export class RelatedMetricsScene extends SceneObjectBase<RelatedMetricsSceneStat
...
     variables: css({
       display: 'none',
     }),
+    searchSticky: css({
+      margin: theme.spacing(1, 0, 1.5, 0),
+      position: 'sticky',
+      background: getAppBackgroundColor(theme, trail),
+      zIndex: 10,
+      paddingBottom: theme.spacing(1),
+      top: `calc(var(--app-controls-height, 0px) + ${headerHeight}px + var(--action-bar-height, 0px))`,
```
shelldandy, line 174: I liked this a lot!

PR #645, opened 2025-08-21
perf(*): Improve native histograms detection

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR is an attempt at improving the performance related to native histogram detection:
1. it eliminates an extra query to fetch the classic histograms and
2. it uses more efficient data structures ([Sets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) and [Maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)).

### 📖 Summary of the changes

Besides changing the data structures and avoiding the extra query, some parts have been refactored to modernize/simplify the code. See diff tab for specific comments.

### 🧪 How to test?

- The build should pass


src/DataTrail.tsx
```diff
@@ -192,7 +190,7 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
   private _addingFilterWithoutReportingInteraction = false;
   private datasourceHelper = new MetricDatasourceHelper(this);
 
-  public getMetricMetadata(metric?: string) {
+  public getMetricMetadata(metric: string) {
```
grafakus, line 193: Strengthen the TS type as it's always called with a metric.

src/DataTrail.tsx
```diff
@@ -216,28 +214,18 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
...
-  }
-
   private resetNativeHistograms() {
     this.setState({
       histogramsLoaded: false,
-      nativeHistograms: [],
     });
   }
 
-  public getCurrentMetricMetadata() {
```
grafakus, line 237: Unused

PR #647, opened 2025-08-22
perf(*): Improve Native Histograms detection

grafakus:
### ✨ Description

**Related issue(s):** follow-up of https://github.com/grafana/metrics-drilldown/pull/645

This PR aims at improving how the app determines if a metric is a native histogram or not. It achieves it by removing the extra request made to fetch all the metrics in favour of listening to the `MetricsVariable` changes. 

The strategy works well in the `MetricsReducer` because, anyway, the app needs the options of `MetricsVariable` to be able to render the metrics list.

When the user lands directly on the `MetricScene`, there's some extra loading time as the app now has to wait for `MetricsVariable` to load its options. The "reward" is that, for native histograms,  the main viz will directly render the correct panel type.

### 📖 Summary of the changes

To make things work:
1. `MetricsVariable` has been moved to `DataTrail` so it always loads its options regardless of where the user lands.
2. `MetricDatasourceHelper` has been updated to listen to any data source change and/or any `MetricsVariable` change and to react accordingly by updating its internal data.
3. the `isNativeHistogram` method has been changed to an async method so that its callers can wait for the correct result before rendering panels.
4. `isNativeHistogram`, when called, ensures that all the metrics metadata are fetched. If the fetching is in progress, no extra request is made. When the fetch results are available, it is cached for future calls.

Note: to continue improving the performance of the app, we should find a way to avoid having to fetch _all_ metrics metadata and fetch only the metadata of a single metric. There's an endpoint in Prometheus to do so, do we have a runtime datasource method?

### 🧪 How to test?

- The build should pass

### DIFF
--- a/src/DataTrail.tsx
+++ b/src/DataTrail.tsx
@@ -15,7 +15,6 @@ import {
   SceneVariableSet,
   ScopesVariable,
   UrlSyncContextProvider,
-  VariableDependencyConfig,
   VariableValueSelectors,
   type SceneComponentProps,
   type SceneObject,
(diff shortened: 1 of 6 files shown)


src/DataTrail.tsx
```diff
@@ -63,20 +62,58 @@ export interface DataTrailState extends SceneObjectState {
   // Synced with url
   metric?: string;
   metricSearch?: string;
-
```
grafakus, line 66: As `DataTrail` doesn't own the native histograms data, it makes sense to remove this data from its state.

bohandley (approved):
This looks great to me! This implementation is much more graceful and reduces the calls for all the metrics and the metadata. I really like the idea of eventually doing a call for a single metric + [metadata](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-target-metadata) when landing on the Metric Scene page. Thank you Marc!

Issue #418, opened 2025-05-13
[Breakdown tab] Use the same label selector as the Profiles & Traces apps


grafakus:
The label selector automatically expands horizontal when space is available on wider screen.

Small screens:
<img width="231" alt="Image" src="https://github.com/user-attachments/assets/1f8d1ef1-402a-4752-9f8d-11134c070249" />

Large screens:
<img width="1062" alt="Image" src="https://github.com/user-attachments/assets/98045085-517b-4b83-a7b3-4034ee9ce1d9" />

**Note:**

Use the latest version of the component in Traces drilldown (which behaves better)

shelldandy:
Setting this to blocked while i work on: https://github.com/grafana/traces-drilldown/issues/548

PR #649, opened 2025-08-22
chore: responsive group-by in breakdown tab

shelldandy:
Fixes: https://github.com/grafana/metrics-drilldown/issues/418

image

image


src/Breakdown/LabelBreakdownScene.tsx
```diff
@@ -69,17 +72,32 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
...
+      throw new Error('ResponsiveGroupBySelector not initialized');
+    }
+    return responsiveSelector;
+  }
+
   public static readonly Component = ({ model }: SceneComponentProps<LabelBreakdownScene>) => {
     const styles = useStyles2(getStyles);
     const { body } = model.useState();
     const groupByVariable = model.getVariable();
+    const useResponsive = true;
```
shelldandy, line 87: quick switch to test back and forth. should i remove the old component now or as a follow-up PR

grafakus (changes requested):
Why didn't we use the same component as the Traces/Profiles Drilldown app: https://github.com/grafana/traces-drilldown/blob/main/src/components/Explore/GroupBySelector.tsx, as mentioned in the issue?

The idea was to reuse code so that we can leverage what other apps have been doing and share what works across apps in the Drilldown ecosystem.

Did you bump into some issues when integrating it?

And I'm afraid that the approach taken in this PR does not work.

Here, the "Group by" selector is not responsive and the  "View" switch is not aligned:

image

Same after selecting a label:

image

shelldandy:
Closing this one in favor of the newer PR with the refactored code from traces drill down

PR #652, opened 2025-08-25
fix(NativeHistograms): Fix queries

grafakus:
### ✨ Description

**Related issue(s):** `-`

The native histograms queries were incorrect. This PR fixes them. One symptom is that more percentiles appeared in the UI:

| Before | After |
|  ---   |  ---  |
| image | image |

Additionally, this PR filters out the "le" label from the "Breakdown" tab, as it represents the upper bound of each classic histogram bucket and is not meant to directly used.

### 📖 Summary of the changes

Just adding the missing `sum` function, as described in https://grafana.com/docs/mimir/latest/visualize/native-histograms/#quantiles.

### 🧪 How to test?

- The build should pass


bohandley:
@NWRichmond yes! This looks right on. I believe these are the correct functions for the types of histograms.

PR #657, opened 2025-08-27
fix(*): Fix queries for label values containing special chars (like the equal sign)

grafakus:
### ✨ Description

**Related issue(s):** follow-up of https://github.com/grafana/metrics-drilldown/pull/623

The linked PR was a quick hack to prevent incorrect queries to be made when some label values contained the equal sign.

This PR solves completely the problem.

### 📖 Summary of the changes

We use the [`:raw` variable formatter](https://grafana.com/docs/grafana/latest/dashboards/variables/variable-syntax/#raw) and let Scenes interpolate and escape the value for us.

### 🧪 How to test?

- The build should pass

### DIFF
--- a/src/DataTrail.tsx
+++ b/src/DataTrail.tsx
@@ -433,7 +433,7 @@ function getVariableSet(initialDS?: string, metric?: string, initialFilters?: Ad
           filters
             .filter((filter) => filter.key !== '__name__')
             // eslint-disable-next-line sonarjs/no-nested-template-literals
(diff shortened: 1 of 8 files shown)


bohandley (commented):
Nice! I wonder though, will this affect when a user has selected multiple values and we get the , instead of the |?

shelldandy (approved):

bohandley:
Good point! We do not have the functionality for a user to select multiple values in our app. To do so, a user need to manually type the values with the `|` delimiter. 

LGTM!

PR #658, opened 2025-08-27
feat(NativeHistograms): Improve detection

grafakus:
### ✨ Description

**Related issue(s):** [this related PR](https://github.com/grafana/metrics-drilldown/pull/660) is based on this branch.

This PR is an attempt at improving how the app detects if a metric is a native histogram or not.

It's achieved by combining 2 things. On one side, **improving how metadata is fetched and cached**, and on the other side, **inspecting the data received by the panel to identify histogram data**.

**A. Fetching metadata**

- The app tries to fetch and cache all the metadata when the user lands on the page. It's generally fast but unreliable. Indeed, I noticed that it can take quite some time for all the metadata to be returned by the data source (more than the first Prometheus scrape interval). This is why...
- ...the app also tries to fetch each individual metric metadata on-demand, whenever it does not exist in its cache. By doing so, we give a chance to the data source to return it during the user journey across the app.

**B. Detecting the data frame type**

When the app couldn't decide if a metric is a native histogram or not, it tries to inspect the first data frame received, and if it's type is "heatmap-cells", it updates the panel type accordingly, which re-renders the panel as a heatmap, without re-fetching any data.

This process is discarded if the metric has already been identified as a classic histogram or a status up/down one. It's been working quite well in my experiments ; we should double-check with the owners of the Prometheus data source to see how reliable is the approach.

**C. Misc**

A basic Prometheus build info has been added to the plugin info dropdown. It's clickable:

| Prometheus | Mimir |
|  ---   |  ---  |
| image | image|

### 📖 Summary of the changes

Even though there are ~60 files changed in this PR, The most important changes are located  in `src/helpers/MetricDatasourceHelper.ts` and `src/GmdVizPanel/GmdVizPanel.tsx` ;)

Just a note here... Because the E2E tests for the `MetricsReducer` started to become messy, I've split them into 2 files:
- `e2e/tests/metrics-reducer/metrics-reducer-panel-types.spec.ts`and
- `e2e/tests/metrics-reducer/metrics-reducer-view.spec.ts`

The corresponding screenshots have moved too.

### 🧪 How to test?

- The tests have been updated (unit + E2E), the build should pass but I'd recommend to also test it locally.

### DIFF
--- a/src/WingmanDataTrail/ListControls/LayoutSwitcher.tsx
+++ b/src/WingmanDataTrail/ListControls/LayoutSwitcher.tsx
@@ -8,6 +8,8 @@ import {
 import { RadioButtonGroup } from '@grafana/ui';
 import React from 'react';
 
+import { reportExploreMetrics } from 'interactions';
+
 export enum LayoutType {
   GRID = 'grid',
   ROWS = 'rows',
@@ -67,6 +69,7 @@ export class LayoutSwitcher extends SceneObjectBase<LayoutSwitcherState> {
   }
 
   private onChange = (layout: LayoutType) => {
+    reportExploreMetrics('layout_changed', { layout });
     this.setState({ layout });
   };
 
--- a/src/interactions.ts
+++ b/src/interactions.ts
@@ -149,6 +149,8 @@ type Interactions = {
   exposed_component_viewed: {
     component: SnakeCase<ExposedComponentName>;
   };
+  // User selects a different layout (grid/rows/single)
+  layout_changed: { layout: LayoutType };
   // User changes the panel type for a histogram metric (e.g., from heatmap to percentiles)
   histogram_panel_type_changed: { panelType: PanelType };
   // App migrated some legacy user prefs (see src/UserPreferences/userStorage.ts)


e2e/fixtures/views/DrilldownView.ts
```diff
@@ -47,6 +47,10 @@ export class DrilldownView {
     return this.page.goBack();
   }
 
+  reload() {
```
grafakus, line 50: It's a leftover from some local experiments. It's non-critical for this PR.

e2e/tests/metrics-reducer/metrics-reducer-panel-types.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    await metricsReducerView.quickSearch.enterText(searchText);
+    await metricsReducerView.assertMetricsList();
+
+    await expectScreenshotInCurrentGrafanaVersion(
+      metricsReducerView.getMetricsList(),
+      'metric-list-with-all-types.png'
+    );
+
+    // same but let's block the metadata fetching
+    await metricsReducerView.route('**/api/datasources/uid/*/resources/api/v1/metadata*', async (route) => {
```
grafakus, line 61: We make sure the approach of inspecting the data frame received works when no metadata is available.

src/helpers/MetricDatasourceHelper.ts
```diff
@@ -85,24 +87,26 @@ export class MetricDatasourceHelper {
...
-  }
 
-  private initializeClassicHistograms(metricsVariableOptions: VariableValueOption[]) {
-    this.classicHistograms = new Set();
+    this.cache = {
+      metadata: new Map(),
+      classicHistograms: new Set(),
+    };
+
+    this.fetchMetricsMetadata().catch(() => {});
```
grafakus, line 101: We fetch all the metadata on each `reset`call:
- When `init` is called (the user lands on the page and the trail is created)
- When the datasource changes

src/test/mocks/datasource.ts
```diff
@@ -37,6 +38,10 @@ class MockDataSource implements DataSourceApi {
       queryLabelKeys: jest.fn(async () => []),
       // eslint-disable-next-line no-unused-vars
       fetchLabelValues: jest.fn(async (_: any) => []),
+      request: jest.fn(async () => ({})),
+    };
+    this.variables = {
+      getType: () => VariableSupportType.Datasource,
```
grafakus, line 44: It prevented some errors to appear in the terminal, but there are more related to `src/DataTrail.test.tsx`, we should look into it in the near future. Less noise === happier dev life

bohandley (approved):
Nice work here! I love the check of data response. This is something I always forget about scenes is that we can get the full lifecycle of the data and change it even after we run a query.

bohandley (approved):
Re-reviewing 🚀

PR #659, opened 2025-08-28
fix(*): Prevent flickering

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR to prevent flicker in the `MetricsReducer` (the list was rendered twice in a row) and the `MetricScene` (the action bar was missing when landing on the page and it would then appear):

| Before | After |
|  ---   |  ---  |
|  https://github.com/user-attachments/assets/abe7d574-70dd-42cf-93a8-4687bd551b09 | https://github.com/user-attachments/assets/c7f61298-aa50-48ff-8f25-2bcbac429092 |

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

`-`


yangkb09 (approved):
thank you for this!!

PR #662, opened 2025-08-29
fix(links): escape value delimiters the scenes way

bohandley:
### ✨ Description

**Related issue(s):** 

When a label value contains `|` this conflicts with scenes URL filter variables that already use `|` as a delimiter between the label name, operator and label value.

Escalation https://github.com/grafana/support-escalations/issues/18160

Screenshot 2025-08-29 at 3 25 31 PM
Screenshot 2025-08-29 at 3 25 46 PM
Screenshot 2025-08-29 at 3 25 55 PM

When a user navigates from dashboard to metrics drilldown, they may have a _special_ value that needs escaped.

Example: `job=~"oteldemo/(kafka|service)"`

The normal special characters are escaped correctly, but the `|` character is special because it is already used as a delimiter in scenes URL variables. `job=~"oteldemo/(kafka|service)` becomes `job|=~|oteldemo/(kafka__gfp__service)`

But what is this `__gfp__` character?? This is the scenes way of escaping delimiters in values that conflict with other scenes delimiters. 

https://github.com/grafana/scenes/blob/39124e94aa134e6ba1ce47680d3037a4193f3c13/packages/scenes/src/variables/utils.ts#L164-L180

We were not doing this in the `links.ts` file and this is a fix for that.

### 📖 Summary of the changes

Escape Prometheus label value character `|` the [scenes way](https://github.com/grafana/scenes/blob/39124e94aa134e6ba1ce47680d3037a4193f3c13/packages/scenes/src/variables/utils.ts#L164-L180)

### 🧪 How to test?

1. Create a dashboard panel with a prometheus data source
2. Create a query for the panel that has a `|`, for example, `{job=~"grafana|mysql"}` (note the regex operator `=~`)
3. Save your dashboard and exit edit mode. 
4. Click the menu item in the panel to find the "open in metrics drilldown" link
Screenshot 2025-08-29 at 3 00 07 PM
5. Open in metrics drilldown and see that the labels are applied correctly.


bohandley:
Going to have to custom copy the Scenes function for now, it is blowing up the bundle entry point size.

I need to add an issue to [export this function](https://github.com/grafana/scenes/blob/39124e94aa134e6ba1ce47680d3037a4193f3c13/packages/scenes/src/variables/utils.ts#L164-L180) from Scenes.

bohandley:
> Going to have to custom copy the Scenes function for now, it is blowing up the bundle entry point size.
> 
> I need to make a PR to [export this function](https://github.com/grafana/scenes/blob/39124e94aa134e6ba1ce47680d3037a4193f3c13/packages/scenes/src/variables/utils.ts#L164-L180) from Scenes.

Here is the PR to export the function from Scenes https://github.com/grafana/scenes/pull/1230

src/extensions/links.ts
```diff
@@ -349,3 +349,13 @@ function prometheusSpecialRegexEscape<T>(value: T) {
     .replace(/\\/g, '\\\\\\\\') // escape backslashes
     .replace(/[$^*{}\[\]+?.()|]/g, '\\\\$&'); // escape regex metacharacters
 }
+
+// Need to export this function from scenes because importing scenesUtils is increasing the bundle entry point size by 522.51kB
+export function escapeUrlPipeDelimiters(value: string | undefined): string {
```
bohandley, line 354: https://github.com/grafana/scenes/pull/1230 to export this function

NWRichmond: Thank you for that issue, and the helpful code comment!

NWRichmond (approved):
Very nice! Thank you for this fix, @bohandley 🤝 

## Local testing results

### On `main`

https://github.com/user-attachments/assets/48c6a2b8-aa11-4110-aee0-fb9d5e90f99b

### On `bohandley/esc-18160/dash-link-filter-values-broken`

https://github.com/user-attachments/assets/f278dec2-ab03-4fcd-a22c-5a97faeff4c7

PR #665, opened 2025-08-30
refactor(links): lazy load lezer parser in links, reduce bundle size

bohandley:
### ✨ Description

**Related issue(s):** From discussion in [Design Doc: Drilldown extension point navigation standard practices](https://docs.google.com/document/d/1qlh_sj25s0Y52hx32dX9ypbT43-_vuPQFHdL0GRRTew/edit?tab=t.0) [from @svennergr](https://docs.google.com/document/d/1qlh_sj25s0Y52hx32dX9ypbT43-_vuPQFHdL0GRRTew/edit?disco=AAABoCBZblM).

This should lazy load the lezer parser and reduce the entry point size.

### 📖 Summary of the changes

1. Move away from using the synchronous configure function for building the URL path
2. Use the `async onClick` function to build the URL path
3. In the async function, lazy load the promql lezer parser
4. Update tests

### 🧪 How to test?

- The tests should pass
- The bundle size should shrink ~50kB
- You may poke around in dashboards, explore query editor or alerting rule editor


src/exposedComponents/LabelBreakdown/LabelBreakdown.tsx
NWRichmond, line 1: If we were using [server components](https://react.dev/reference/rsc/server-components), we could leverage async/await like this. But until we can use server components in a Grafana app plugin, we'll need to take a different approach.

src/exposedComponents/LabelBreakdown/LazyLabelBreakdown.tsx
NWRichmond, line 1: I think we'll need to find a different approach here. When using this with `asserts-app` locally, I'm seeing [this error](https://react.dev/errors/321?invariant=321):

> Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
> 1. You might have mismatching versions of React and the renderer (such as React DOM)
> 2. You might be breaking the Rules of Hooks
> 3. You might have more than one copy of React in the same app
See https://react.dev/link/invalid-hook-call for tips about how to debug and fix this problem.

src/extensions/links.ts
```diff
@@ -41,7 +40,14 @@ export const linkConfigs: Array<PluginExtensionAddedLinkConfig<PluginExtensionPa
...
     ],
-    configure: configureDrilldownLink,
+    configure: configureDrilldownLinkForAsync,
+    onClick: async (_, helpers) => {
+      const context = helpers.context;
+      const url = await buildDrilldownUrlAsync(context);
+      if (url) {
+        window.location.href = url;
+      }
+    },
```
NWRichmond, line 50: Cool! I didn't know that `configure` and `onClick` could be used together. TIL :)

NWRichmond (commented):
Off to a great start!

We just need to figure out the pattern for lazy-loaded exposed components. I'd be happy to help with this, if you'd like.

NWRichmond:
@bohandley I tried an effect-driven approach to using the async `parsePromQLQuery` in `src/exposedComponents/LabelBreakdown/LabelBreakdown.tsx`, and it's working nicely! I'm happy to push to this branch or open a PR.

NWRichmond (dismissed):
Excellent!

The `LabelBreakdown.tsx` exposed component is working nicely in the Asserts app locally.

I'm very excited about the huge bundle size savings. This sets a great precedent for other apps that use hefty parsers to construct links.

PR #667, opened 2025-09-01
feat(RelatedLogsScene): enable log panel options

matyax:
This PR follows up a recent conversation with @NWRichmond, where we talked about related logs within Metrics. There's a lot of different ideas that can be implemented, from embedding Logs Drilldown, to add metric-relevant filters from Log Details.

Here, we begin with the lowest hanging fruit, which is enabling more options in the Metrics' Logs Panel.

Changes:
- Added heights to allow the panel to use the available space
- Enabled visualization options.

Before:

Before

After:

After


src/RelatedLogs/RelatedLogsScene.tsx
```diff
@@ -133,9 +135,16 @@ export class RelatedLogsScene extends SceneObjectBase<RelatedLogsSceneState> {
     }
 
     // Set up UI for logs panel
-    const logsPanelContainer = sceneGraph.findByKeyAndType(this, LOGS_PANEL_CONTAINER_KEY, SceneCSSGridItem);
+    const logsPanelContainer = sceneGraph.findByKeyAndType(this, LOGS_PANEL_CONTAINER_KEY, SceneFlexItem);
     logsPanelContainer.setState({
-      body: PanelBuilders.logs().setTitle('Logs').setData(this._queryRunner).build(),
+      body: PanelBuilders.logs()
+        .setTitle('Logs')
```
matyax, line 141: Here can be seen how we pass different callbacks to the visualization that enable interactions within log details:

https://github.com/grafana/logs-drilldown/blob/main/src/Components/ServiceScene/LogsPanelScene.tsx#L335

For example, what happens when these buttons are clicked:

imagen

matyax:
Looking forward to see what you think!

grafakus (approved):
Nice, thank you @matyax !

The only thing that bothers me a bit is the double-scroll (the logs panel should occupy all the available visible space):

https://github.com/user-attachments/assets/f1199e3c-0069-43ee-a2b7-b714849a63d0

But we don't have to fix it in this PR. I think @yangkb09 is working on the layout of the `MetricScene` so we can tackle it later.

matyax:
I hear you @grafakus ! The limiting factor for height are the controls:

imagen

I just updated the min height to allow for a smaller panel.

PR #673, opened 2025-09-03
Chore/group by take 2

shelldandy:
Fixes: https://github.com/grafana/metrics-drilldown/issues/418

Had to refactor the component in traces-drilldown first in: https://github.com/grafana/traces-drilldown/pull/549

With that done, i could then actually just copy and paste the folder in here

### DIFF
--- a/src/Breakdown/GroupBySelector/GroupBySelector.tsx
+++ b/src/Breakdown/GroupBySelector/GroupBySelector.tsx
@@ -1,0 +1,209 @@
+import { css } from '@emotion/css';
+import { useResizeObserver } from '@react-aria/utils';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
+
+import { GrafanaTheme2 } from '@grafana/data';
+import { Combobox, RadioButtonGroup, useStyles2, useTheme2, Field } from '@grafana/ui';
+
+import {
+  GroupBySelectorProps,
+  FilterContext,
+  DomainConfig,
+} from './types';
+import {
+  processRadioAttributes,
+  filteredOptions,
+  getModifiedSelectOptions,
+  createDefaultGroupBySelectorConfig,
+  mergeConfigurations,
+} from './utils';
+
+const DEFAULT_ADDITIONAL_WIDTH_PER_ITEM = 40;
+const DEFAULT_WIDTH_OF_OTHER_ATTRIBUTES = 180;
+const DEFAULT_ALL_OPTION = 'All';
+
+export function GroupBySelector({
+  // Core props
+  options,
+  radioAttributes,
+  value,
+  onChange,
+  showAll = false,
+
+  // State data
+  filters = [],
+  currentMetric,
+  initialGroupBy,
+
+  // Display configuration
+  attributePrefixes = {},
+  fieldLabel = 'Group by',
+  selectPlaceholder = 'Other attributes',
+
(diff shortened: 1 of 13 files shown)


grafakus:
Hey Miguel, it still does not work, the quick search does not grow:

image

src/Breakdown/GroupBySelector/utils.ts
```diff
@@ -0,0 +1,222 @@
...
+  return {
+    attributePrefixes: {},
+    filteringRules: {
+      excludeFilteredFromRadio: true,
+      // Custom filter to exclude histogram bucket labels like GroupByVariable does
+      customAttributeFilter: (attribute: string) => {
+        return attribute !== 'le';
+      }
+    },
+    ignoredAttributes: ['__name__', 'timestamp', 'le'],
```
grafakus, line 186: It's been taken care of in https://github.com/grafana/metrics-drilldown/pull/679

src/Breakdown/GroupBySelector/GroupBySelector.tsx
```diff
@@ -0,0 +1,218 @@
...
+    onResize: () => {
+      const element = controlsContainer.current;
+      if (element && config.layoutConfig.enableResponsiveRadioButtons) {
+        setAvailableWidth(element.clientWidth);
+      }
+    },
+  });
+
+  // Create filter context
+  const filterContext: FilterContext = useMemo(() => ({
```
grafakus, line 90: `useMemo` should be [used to cache expensive calculations](https://react.dev/reference/react/useMemo#usage), which is not the case here.

src/Breakdown/LabelBreakdownScene.tsx
```diff
@@ -74,15 +76,122 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
...
+            operator: f.operator,
+            value: f.value
+          }))
+        : [],
+      [filtersVariable]
+    );
+
+    // Define common Prometheus metric labels for radio buttons
+    // Ordered by importance - most common labels first
+    const commonPrometheusLabels = useMemo(() => [
```
grafakus, line 99: Same as my previous comment. It does not make sense to use `useMemo`.

src/Breakdown/LabelBreakdownScene.tsx
```diff
@@ -74,15 +76,122 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
     const { body } = model.useState();
     const groupByVariable = model.getVariable();
 
+    // Extract state manually from scene graph (Phase 2: Direct Migration)
```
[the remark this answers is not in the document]
shelldandy: my bad, this was leftover from the migration process of the PR in earlier commits

grafakus:
> @grafakus we need to reserve some space for the groupby component so it can do its calculations of when to show/hide the dropdown.

Yes I understand, but the idea of adding a responsive component is to dynamically adapt to all the available space. But it does not happen in this PR: 

image

Hence my suggestion to pair to look into it. Unless you already want try a better approach. Thank you!

NWRichmond (commented):
I'm excited by the feature, but there are some things that aren't quite there yet. I'm happy to provide more detail, meet, etc. - let me know :)

PR #674, opened 2025-09-04
fix(RelatedLogs): Fix fatal error when no logs are found

grafakus:
### ✨ Description

**Related issue(s):** `-`

Just a quick fix when the user goes to the `MetricScene`, clicks on the "Related logs" tab and no logs are found:

image

### 📖 Summary of the changes

A call to `sceneGraph.findByKeyAndType` with the wrong type of object was causing a runtime error.

### 🧪 How to test?

- A simple E2E test has been added in this PR ; the build should pass

**Note:** in the near future, we should setup and start testing with a Loki data source

### DIFF
--- a/src/RelatedLogs/RelatedLogsScene.tsx
+++ b/src/RelatedLogs/RelatedLogsScene.tsx
@@ -2,7 +2,6 @@ import { LoadingState } from '@grafana/data';
 import {
   CustomVariable,
   PanelBuilders,
-  SceneCSSGridItem,
   SceneFlexItem,
   SceneFlexLayout,
   sceneGraph,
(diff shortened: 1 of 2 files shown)


src/RelatedLogs/RelatedLogsScene.tsx
```diff
@@ -86,7 +85,7 @@ export class RelatedLogsScene extends SceneObjectBase<RelatedLogsSceneState> {
   }
 
   private showNoLogsFound() {
-    const logsPanelContainer = sceneGraph.findByKeyAndType(this, LOGS_PANEL_CONTAINER_KEY, SceneCSSGridItem);
+    const logsPanelContainer = sceneGraph.findByKeyAndType(this, LOGS_PANEL_CONTAINER_KEY, SceneFlexItem);
```
grafakus, line 88: See https://github.com/grafana/metrics-drilldown/blob/main/src/RelatedLogs/RelatedLogsScene.tsx#L57-L60

matyax (approved):
Good find!

PR #676, opened 2025-09-04
perf(GmdVizPanel): Faster first render + small refatorings

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR decreases the time to first render of the `GmdVizPanel` by determing if the metric is a native histogram or not **after** rendering the component. Indeed it's an async process that may require a call to the backend API.

Locally, some tests have shown a decrease of 100ms between the activation of the Scene object and the 1st render of the component.

Additionally:
1. prior to this PR, the metric type that was tracked when the user opened the config Prometheus form did not take into consideration if the metric was a native histogram or not. The same was happening when the config was applied or restored,
2. this PR adds matchers in `src/GmdVizPanel/matchers`: `getPanelTypeForMetric` and `getPanelTypeForMetricSync`
3. and this PR also adds some missing unit tests

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- The build should pass

### DIFF
--- a/src/DataTrail.tsx
+++ b/src/DataTrail.tsx
@@ -32,7 +32,7 @@ import { EventApplyPanelConfig } from 'GmdVizPanel/components/ConfigurePanelForm
 import { EventCancelConfigurePanel } from 'GmdVizPanel/components/ConfigurePanelForm/EventCancelConfigurePanel';
 import { EventConfigurePanel } from 'GmdVizPanel/components/EventConfigurePanel';
 import { GmdVizPanel } from 'GmdVizPanel/GmdVizPanel';
-import { getMetricType } from 'GmdVizPanel/matchers/getMetricType';
+import { getMetricType, getMetricTypeSync } from 'GmdVizPanel/matchers/getMetricType';
 import { MetricsDrilldownDataSourceVariable } from 'MetricsDrilldownDataSourceVariable';
 import { PluginInfo } from 'PluginInfo/PluginInfo';
 import { displaySuccess } from 'WingmanDataTrail/helpers/displayStatus';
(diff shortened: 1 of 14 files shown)


src/DataTrail.tsx
```diff
@@ -178,13 +178,17 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
     this.subscribeToEvent(EventConfigurePanel, async (event) => {
       const { metric } = event.payload;
 
-      reportExploreMetrics('configure_panel_opened', { metricType: getMetricType(metric) });
+      getMetricType(metric, this)
```
grafakus, line 181: Non blocking ; with fallback.

src/GmdVizPanel/GmdVizPanel.tsx
```diff
@@ -145,6 +145,11 @@ export class GmdVizPanel extends SceneObjectBase<GmdVizPanelState> {
   private async onActivate(discardPanelTypeUpdates: boolean) {
     const { metric, panelConfig } = this.state;
 
+    this.updateBody();
```
grafakus, line 148: The comment above L121 was explaining something that was not happening. My mistake 🤦🏾‍♂️

src/GmdVizPanel/components/ConfigurePanelForm/ConfigurePanelForm.tsx
```diff
@@ -73,8 +73,7 @@ export class ConfigurePanelForm extends SceneObjectBase<ConfigurePanelFormState>
   private async buildBody() {
     const { metric } = this.state;
     const prefConfig = getPreferredConfigForMetric(metric);
-    const isNativeHistogram = await getTrailFor(this).isNativeHistogram(metric);
-    const presets = getConfigPresetsForMetric(metric, isNativeHistogram);
+    const presets = await getConfigPresetsForMetric(metric, getTrailFor(this));
```
grafakus, line 76: Encapsulation FTW

Issue #420, opened 2025-05-13
[MetricScene] Consider moving the "Select new metric", "Open in Explore", "Copy URL" and "Bookmark" buttons higher in the UI


grafakus:
**Current placement:**
<img width="293" alt="Image" src="https://github.com/user-attachments/assets/45eeb9b3-25ac-4d7c-970f-91b21e1b89be" />

**Future placement:** next to the "Settings" button + remove the "Select new metric" button (the browser back button now works properly)

**Related issue with designs:** https://github.com/grafana/grafana/issues/90834

PR #680, opened 2025-09-05
fix(actions): update actions UI

bohandley:
### ✨ Description

**Related issue(s):** https://github.com/grafana/metrics-drilldown/issues/420

This moves the action bar items to the following places:

Select new metric -> top controls
Open in explore -> panel menu item
Copy URL -> metric graph panel menu item
Bookmark -> metric graph header action (similar to Profiles drilldown)

Also, in the metric reducer scene, each panel has a "select" primary action. This has been moved to the far right.

This UI update is the [result of discussion here](https://docs.google.com/document/d/1azkArsmkx-qj8Bhm418_bF33SwUauCFmjcPGqHuE028/edit?tab=t.0) with the Metrics Drilldown squad and @jessover9000 
Here is the [figma board](https://www.figma.com/board/IAATBRDX7d1GnesyaHcmhe/Metrics-Drilldown-Design-Support?node-id=0-1&t=gTEC4LNx2pdFUyXM-0) as well.

Screenshot 2025-09-12 at 12 26 46 AM

The configure and primary action have been switched. Primary action goes in the top right corner.

### 📖 Summary of the changes

1. Moved actions
2. Wrote bookmark action that does not use a hook (is more refactoring needed?)
3. Refactored items

### 🧪 How to test?

1. Try out the actions
  - select a new metric
  - copy url
  - bookmark


bohandley:
If we're happy about this design, I'll regenerate the screenshots :)

src/Menu/PanelMenu.tsx
```diff
@@ -38,50 +42,32 @@ export class PanelMenu extends SceneObjectBase<PanelMenuState> implements VizPan
...
-            };
-          }
-
-          return query;
-        });
-      } catch {}
+      // Check if this is the main metric graph panel
+      // The main panel is created with labelName matching the metric in MetricGraphScene
+      const trail = getTrailFor(this);
+      const isMainGraphPanel = this.state.labelName === trail.state.metric;
```
grafakus, line 48: This is very brittle.

Maybe the easiest and most resilient thing to do instead would be to pass a key to the panel menu [when it's created in MetricGraphScene](https://github.com/grafana/metrics-drilldown/blob/main/src/MetricGraphScene.tsx#L90)? And then, this check would work:

```suggestion
      const isMainGraphPanel = this.state.key === TOPWVIEW_PANEL_MENU_KEY;
```

grafakus: Also, a very simple readability improvement:

```ts
 this.addActivationHandler(this.onActivate.bind(this));
```

src/controls/SelectNewMetricButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+    reportExploreMetrics('selected_metric_action_clicked', { action: 'unselect' });
+    trail.publishEvent(new MetricSelectedEvent({}));
+  };
+
+  static readonly Component = ({ model }: SceneComponentProps<SelectNewMetricButton>) => {
+    const trail = getTrailFor(model);
+    const { topScene, embedded } = trail.useState();
+
+    // Only show the button when a metric is selected (topScene is MetricScene)
+    const isButtonVisible = topScene instanceof MetricScene;
```
grafakus, line 32: Deciding if the button is visible or not should not be the responsibility of the button itself but of `DataTrail`. 

By doing so, the code here would be a bit simpler and the component would not re-render every time the state of `DataTrail` changes (because of L29).

e2e/tests/metric-scene-view.spec.ts
```diff
@@ -144,7 +144,9 @@ test.describe('Metric Scene view', () => {
       });
 
       test('No related logs found', async ({ metricSceneView }) => {
-        await expect(metricSceneView.getTabContent().getByText('No related logs found')).toBeVisible();
+        await expect(
```
bohandley, line 147: This was flaky and preventing screenshot regeneration.

docker-compose.yaml
```diff
@@ -70,5 +70,5 @@ services:
     environment:
       GRAFANA_PORT: ${GRAFANA_PORT:-3001}
       GRAFANA_SCOPES_PORT: ${GRAFANA_SCOPES_PORT:-3002}
-      GRAFANA_IMAGE: ${GRAFANA_VERSION:-grafana-enterprise}
+      GRAFANA_IMAGE: ${GRAFANA_IMAGE:-grafana-enterprise}
```
bohandley, line 73: This was the cause for the grafana version errors in the screen shot regeneration script 🤯

bohandley (commented):
There is one more e2e test issue with v 12.0.3 and I think it is with scopes. I'm close!

docker-compose.yaml
```diff
@@ -67,8 +79,15 @@ services:
       - ./src:/app/src
     command: ${PLAYWRIGHT_ARGS:-}
     profiles: [playwright]
+    depends_on:
+      grafana-scopes-gmd:
```
bohandley, line 83: This scope test was failing in CI but passing locally 🫨 Health check added

e2e/tests/metrics-reducer/metrics-reducer-view.spec.ts-snapshots/metrics-reducer-group-by-label-after-clear-filter-chromium-linux.png
grafakus, line 1: @bohandley  The new screenshot looks weird: there's a lot of space between the panels in the grid

src/GmdVizPanel/components/BookmarkHeaderAction.tsx
```diff
@@ -0,0 +1,114 @@
...
+    });
+  }
+
+  private isCurrentStateBookmarked(): boolean {
+    try {
+      const trail = getTrailFor(this);
+      const currentUrlState = sceneUtils.getUrlState(trail);
+      const currentKey = genBookmarkKey(currentUrlState);
+      const bookmarksFromStorage = userStorage.getItem(PREF_KEYS.BOOKMARKS) || [];
+      return bookmarksFromStorage.some((b: any) => genBookmarkKey(b.urlValues) === currentKey);
```
grafakus, line 43: Just a TS nit, if we export the proper type from [useBookmark.ts](https://github.com/grafana/metrics-drilldown/blob/main/src/bookmarks/useBookmarks.ts#L20): 

```suggestion
      return bookmarksFromStorage.some((b: BookmarkFromStorage) => genBookmarkKey(b.urlValues) === currentKey);
```

src/GmdVizPanel/components/BookmarkHeaderAction.tsx
```diff
@@ -0,0 +1,114 @@
...
+      return false;
+    }
+  }
+
+  public onClick = () => {
+    const trail = getTrailFor(this);
+    const currentUrlState = sceneUtils.getUrlState(trail);
+    const currentKey = genBookmarkKey(currentUrlState);
+    const bookmarksFromStorage = userStorage.getItem(PREF_KEYS.BOOKMARKS) || [];
+    const isCurrentlyBookmarked = bookmarksFromStorage.some((b: any) => genBookmarkKey(b.urlValues) === currentKey);
```
grafakus, line 54:
```suggestion
    const isCurrentlyBookmarked = this.state. isBookmarked;
```

src/MetricGraphScene.tsx
```diff
@@ -55,9 +56,9 @@ export class MetricGraphScene extends SceneObjectBase<MetricGraphSceneState> {
               panelOptions: {
                 height: PANEL_HEIGHT.XL,
                 headerActions: isClassicHistogramMetric(metric)
-                  ? () => [new GmdVizPanelVariantSelector({ metric }), new ConfigurePanelAction({ metric })]
-                  : () => [new ConfigurePanelAction({ metric })],
-                menu: () => new PanelMenu({ labelName: metric }),
+                  ? () => [new GmdVizPanelVariantSelector({ metric }), new ConfigurePanelAction({ metric }), new BookmarkHeaderAction({ metric })]
+                  : () => [new ConfigurePanelAction({ metric }), new BookmarkHeaderAction({ metric })],
+                menu: () => new PanelMenu({ labelName: metric, key: 'TOPWVIEW_PANEL_MENU_KEY' }),
```
grafakus, line 61: Out of curiosity, why not using a constant as it's done for with `TOPVIEW_PANEL_KEY`?

docker-compose.yaml
grafakus: Nice catch!

docker-compose.yaml
grafakus: Nice to go with the Docker way!

Can we remove the [custom webServer hack](https://github.com/grafana/metrics-drilldown/blob/main/e2e/config/playwright.config.common.ts#L56-L63) from Playwright's config then?

grafakus (commented):
Very very nice job Brendan! 🙌🏾

Just left a couple of suggestions for improvements that should be very easy to tackle

docker-compose.yaml
bohandley: I think so! Can we add that as a fast follow to test it on it's own? I am new to the complexities of setting up the e2e work but I think we can remove the thing from the playwright config. I would want to ping @shelldandy for the follow up to remove this.

src/Menu/actions/ExploreAction.ts
```diff
@@ -0,0 +1,46 @@
...
+      const panelData = sceneGraph.getData(viz).state.data;
+      if (!panelData) {
+        throw new Error('Cannot get link to explore, no panel data found');
+      }
+      const queryRunner = getQueryRunnerFor(viz);
+      const queries = queryRunner?.state.queries ?? [];
+      queries.forEach((query) => {
+        // removing legendFormat to get verbose legend in Explore
+        delete query.legendFormat;
+      });
```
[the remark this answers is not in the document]
bohandley: > 1. modifying the queries doesn't have any effect at all

It does not!

> why modifying the legend anyway?

We modify the legend to respect the default legend in Explore query editor (this is weird, I know). I think this is a design choice to respect the default Explore query editor legend for Prometheus. We are opinionated about Metrics Drilldown legend, and then when someone moved to a query editor, we let that do it's job creating the Explore legend. BUT it is not being reset and this would create a disjointed experience of a different legend.

I will remove it!

bohandley (commented):
Updated the changes, thank you for the review Marc!

Issue #426, opened 2025-05-13
Integrate Metrics Drilldown into Asserts


grafakus:
_TDB_

PR #691, opened 2025-09-09
fix(DataTrail): Set URL search parameters namespace when the app/components are embedded

grafakus:
### ✨ Description

**Related issue(s):** relates to https://github.com/grafana/metrics-drilldown/issues/426

The support for namespacing URL search parameters was added to Scenes in https://github.com/grafana/scenes/pull/1156 but Metrics Drilldown is not using it, so its variables might collide with other variables when the app is embedded. 

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

`-`


NWRichmond (approved):
Thanks!

PR #702, opened 2025-09-16
chore: Responsive GroupBySelector - No observer

shelldandy:
Fixes: https://github.com/grafana/metrics-drilldown/issues/418

PR #705, opened 2025-09-18
chore: Follow-up GroupBySelector

grafakus:
### ✨ Description

**Related issue(s):** follow-up of https://github.com/grafana/metrics-drilldown/pull/702

### 🧪 How to test?

- The build should pass

### DIFF
--- a/e2e/fixtures/views/MetricSceneView.ts
+++ b/e2e/fixtures/views/MetricSceneView.ts
@@ -156,59 +156,56 @@ export class MetricSceneView extends DrilldownView {
   }
 
   async assertDefaultBreadownListControls() {
-    await this.assertLabelDropdown('All');
+    await this.assertLabelSelector('All');
     await expect(this.getLayoutSwitcher()).toBeVisible();
     await this.assertSelectedLayout('Grid');
   }
 
-  getLabelDropdown() {
+  getLabelSelectorContainer() {
     return this.getByTestId('breakdown-label-selector');
   }
 
-  async assertLabelDropdown(optionLabel: string) {
+  async assertLabelSelector(label: string) {
     // Check if it's selected as a radio button
-    const radioOption = this.getLabelDropdown().getByRole('radio', { name: optionLabel });
-    if (await radioOption.isVisible()) {
+    const radioGroup = this.getLabelSelectorContainer().getByRole('radiogroup');
+    if (await radioGroup.isVisible()) {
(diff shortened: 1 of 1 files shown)


e2e/fixtures/views/MetricSceneView.ts
```diff
@@ -156,59 +156,56 @@ export class MetricSceneView extends DrilldownView {
...
-    const radioOption = this.getLabelDropdown().getByRole('radio', { name: label });
-    try {
-      if (await radioOption.isVisible({ timeout: 2000 })) {
-        await radioOption.click();
-        return;
-      }
-    } catch {
-      // Radio button not found or not visible, continue to combobox
+    // Check it's a radio button
+    const radioGroup = this.getLabelSelectorContainer().getByRole('radiogroup');
```
grafakus, line 189: This change is to avoid timeouts so we keep the tests fast. Instead of relying on the options to be visible, which happens only after they are fully loaded, we check the container type (`radiogroup` or `combobox` roles) so the test can act directly after.

src/Breakdown/GroupBySelector/index.ts
grafakus, line 1: https://tkdodo.eu/blog/please-stop-using-barrel-files ;)

grafakus: Especially in this case: not all exports were imported

src/Breakdown/GroupBySelector/GroupBySelector.tsx
```diff
@@ -1,117 +1,70 @@
...
+    </div>
   );
 }
 
 function getStyles() {
   return {
-    selectContainer: css({
-      minWidth: 100,
+    combobox: css({
+      marginLeft: '4px',
```
grafakus, line 67: Without the left margin:

image

src/Breakdown/LabelBreakdownScene.tsx
```diff
@@ -100,12 +95,11 @@ function getStyles(theme: GrafanaTheme2) {
       display: 'flex',
       minHeight: '100%',
       flexDirection: 'column',
-      paddingTop: theme.spacing(1),
     }),
     controls: css({
       flexGrow: 0,
       display: 'flex',
-      gap: theme.spacing(2),
+      gap: theme.spacing(1),
```
grafakus, line 102: Save some width

shelldandy (approved):

Issue #698, opened 2025-09-12
Leverage metric metadata to distinguish metric type


NWRichmond:
## Motivation

Sometimes we encounter metrics that end in `_sum`, which leads us to believe that the metric is a counter, when it is in fact a gauge. This is one example (but might not be the only example) outside of Native Histograms where the [metric metadata provided by the Prometheus data source](https://github.com/grafana/metrics-drilldown/blob/68125a2a6fcb7bf81b0d90d10cdecf98e2990fb9/src/helpers/MetricDatasourceHelper.ts#L143-L166) can help us craft the correct PromQL expression for a particular metric. 

## Scope & approach

For now, let's focus on leveraging the metric metadata to check the `type` when building time series query runner params. Currently, whether or not a rate query is used is determined by the suffix:

https://github.com/grafana/metrics-drilldown/blob/68125a2a6fcb7bf81b0d90d10cdecf98e2990fb9/src/GmdVizPanel/types/timeseries/getTimeseriesQueryRunnerParams.ts#L24

https://github.com/grafana/metrics-drilldown/blob/68125a2a6fcb7bf81b0d90d10cdecf98e2990fb9/src/GmdVizPanel/matchers/isCounterMetric.ts#L1-L3

We can continue to eagerly determine the metric type based on metric name heuristics, but there should be an async check that checks the metric metadata for `metadata.type` and updates the query if the assumed metric type isn't correct.

PR #717, opened 2025-09-19
feat: use metadata to distinguish metric type

bohandley:
### ✨ Description

**Related issue(s):** https://github.com/grafana/metrics-drilldown/issues/698

If a metric has a name that indicates it's a counter, but is actually a gauge, we do not use the rate query for that.

https://github.com/user-attachments/assets/6b9e31df-5c7d-4549-a93a-bf18225cd246

### 📖 Summary of the changes

1. Check when building time series panels
2. Provide async metadata check to see if metadata type == type from metric name
3. Override in panel viz query, rewrite without rate query
4. Run query again 

### 🧪 How to test?

1. Build should pass
2. I made a custom gauge with sum suffix metric in gdev prometheus, but can we do this for our provisioned data sources?
3. If we can make this custom metric, we can test it like my video. Otherwise I will make a PR in grafana to update gdev-prometheus to use this new metric.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -66,6 +66,14 @@ services:
       --config.file=/etc/prometheus/prometheus.yml
       --storage.tsdb.path=/prometheus/data
       --storage.tsdb.retention.time=42y
+    healthcheck:
+      # healthy only when metrics metadata is returned
+      # it's important because the app heavily relies on them to determine how metrics are rendered in the UI (see GmdVizPanel.tsx)
+      test: ['CMD-SHELL', 'wget -qO- http://localhost:9090/api/v1/targets/metadata | grep -Fq ''"data":[{''']
+      interval: 3s
+      timeout: 1s
+      retries: 10
+      start_period: 15s
 
   # e2e testing
   playwright:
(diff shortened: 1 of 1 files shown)


NWRichmond (approved):
Nicely done! By setting a breakpoint in `src/shared/GmdVizPanel/types/timeseries/buildTimeseriesPanel.ts` near this line:

```tsx
const corrected = getTimeseriesQueryRunnerParams({ metric, queryConfig, isRateQueryOverride: isCounter });
```

, I found that `gdev-prometheus` data source's `grafana_live_transport_messages_sent_size` metric is an example of a counter that's improperly categorized as a gauge using metric name heuristics. The metadata is being used to determine the correct type and update the query accordingly.

<details>
  <summary>Expand to see comparison</summary>
  
  ## On `main`

Image

Image

## On `bohandley/use-metadata-to-distinguish-metric-type-698`

Image

Image
</details>

With those test results, this appears to be a great solution for https://github.com/grafana/metrics-drilldown/issues/698. I'll look forward to approving again after the merge conflicts are fixed.

shelldandy (approved):
lgtm!

PR #741, opened 2025-09-26
fix(MetricDatasourceHelper): Fix cache refresh when changing data source

grafakus:
### ✨ Description

**Related issue(s):** `-`

This PR ensures that the cache metadata is _always_ cleared when the data source changes.

The previous code relied on subscribing to `MetricsDrilldownDataSourceVariable` state changes but weirdly, when (e.g.) going to the `MetricScene`, coming back to `MetricsReducer` and then changing the data source, the subscriber function would not be called 🤷🏾‍♂️.

### 📖 Summary of the changes

Instead of subscribing to `MetricsDrilldownDataSourceVariable` state changes, this PR relies on the Scenes variable dependency.

### 🧪 How to test?

Manually, after checking out this PR's branch, by looking at the network tab to see:
- that the new metadata is fetched _once_ when landing
- that the new metadata is properly refetched when changing data source

### DIFF
--- a/src/AppDataTrail/DataTrail.tsx
+++ b/src/AppDataTrail/DataTrail.tsx
@@ -15,6 +15,7 @@ import {
   SceneVariableSet,
   ScopesVariable,
   UrlSyncContextProvider,
+  VariableDependencyConfig,
   VariableValueSelectors,
   type SceneComponentProps,
   type SceneObject,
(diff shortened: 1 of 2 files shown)


NWRichmond (approved):
Great catch, and awesome code comments!

It works as expected:

https://github.com/user-attachments/assets/4664efd2-e5f6-4026-91b7-95f471011e4d

Thanks for this :)

shelldandy (approved):

PR #744, opened 2025-09-29
feat(*): Introduce presets for "_info" metrics

grafakus:
### ✨ Description

**Related issue(s):** relates to https://github.com/grafana/metrics-drilldown/issues/450#issuecomment-3126964781 (idea barely touched) and requested [in Slack today](https://raintank-corp.slack.com/archives/C075BDBTX96/p1759143847962589)

This PR changes the Prometheus function associated with "_info" metrics, from "avg" to "count":

https://github.com/user-attachments/assets/7637a14a-11c7-4c69-95fe-76319ad37c70

Some ideas for the future:
1. As mentioned in https://github.com/grafana/metrics-drilldown/issues/450, the app could add a preset to display a table with label cardinalities instead of a timeseries with "count"
2. Or we could leave a timeseries with "count" and offer a dropdown in the header of the panel to switch between different labels, which would group by that label (think about it as the "percentiles"/"heatmap" switch for histograms in the `MetricScene`)

### 📖 Summary of the changes

The changes are quite simple:

1. a new metric type `info` has been created,
3. a new Prometheus function `count` is now supported and is associated to `info` metrics by default,
4. a new preset for `info` metrics has been created and is displayed in the UI in the config form (currently it's a single preset)

### 🧪 How to test?

- The build with the updated E2E should pass


src/shared/GmdVizPanel/config/presets/getConfigPresetsForMetric.ts
```diff
@@ -24,6 +25,9 @@ export async function getConfigPresetsForMetric(metric: string, dataTrail: DataT
     case 'status-updown':
       return Object.values(DEFAULT_STATUS_UP_DOWN_PRESETS);
 
+    case 'info':
```
shelldandy, line 28: Wouldn't it make more sense to use `count` as well here?

shelldandy (approved):
Just had a question but overall lgtm thanks for this!

src/shared/GmdVizPanel/config/presets/getConfigPresetsForMetric.ts
grafakus: Having said this... Your comment made me realize that there was some naming inconsistency in the preset names, fixed in https://github.com/grafana/metrics-drilldown/pull/744/commits/bc2345d654552841fd51ec27e5718c3eee6ccd1d (thank you!)

PR #745, opened 2025-09-30
feat(Breakdown): Disable __ignore_usage__ label for the main query

grafakus:
### ✨ Description

**Related issue(s):** see [Slack context](https://raintank-corp.slack.com/archives/C075BDBTX96/p1759166711617869)

This PR removes the `__ignore_usage__` label for Breakdown label-values query. For example, for the `instance` label:

| Before | After |
|  ---   |  ---  |
| image | image |

**Motivation:** selecting a label in Breakdown signals user interest. Keeping `__ignore_usage__=""` in the query causes Adaptive Metrics to ignore that query. Ignored queries can keep the selected label aggregated, contradicting user intent.

### 📖 Summary of the changes

See diff tab

### 🧪 How to test?

- The build should pass


shelldandy (approved):

grafakus:
> hmm i still see the `__ignore_usage__` label after checking the query on your branch - am i missing something? 🤔 

The label is removed only in the 1st query that occurs when you select a label (`instance` in the screenshots above). Sorry it could have been clearer in the PR description. I've updated it to clarify this point.

yangkb09:
> > hmm i still see the `__ignore_usage__` label after checking the query on your branch - am i missing something? 🤔
> 
> The label is removed only in the 1st query that occurs when you select a label (`instance` in the screenshots above). Sorry it could have been clearer in the PR description. I've updated it to clarify this point.

i see! ok nice, i see it now after selecting a label
image

PR #747, opened 2025-09-30
chore: cd deploys to dev

shelldandy:
Fixes https://github.com/grafana/metrics-drilldown/issues/750

Adds an action to do CD to dev on each commit from `main`

I couldn't test it from the branch, seems like its a known limitation from the plugins-ci-workflows as such @L2D2Grafana suggested we should just merge and check


.github/workflows/publish-dev.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  CD:
+    name: Deploy plugin
+    uses: grafana/plugin-ci-workflows/.github/workflows/cd.yml@main
+    with:
+      branch: ${{ github.event.inputs.branch }}
+      # When pushing to "main", publish and deploy to "dev" and "ops" (CD). For PRs, skip publishing and deploying (run CI only)
+      environment: dev
+      attestation: true # this guarantees that the plugin was built from the source code provided in the release
+      run-playwright: false
```
shelldandy, line 23: We already have other actions that run playwright for us on `main`

yangkb09 (approved):

PR #765, opened 2025-10-07
fix: init plugin translations for Scenes

NWRichmond:
### ✨ Description

When I try to `npm run dev` on the latest `main`, I encounter this error:

demo scenes translation init issue

After `rm -rf node_modules/ && npm i`, the problem persists. I'm not sure why this is happening all of a sudden, and so far it appears to not be affecting others. So let's treat this PR as a placeholder, in case this is truly an issue outside of my local dev environment that needs solving.

### 📖 Summary of the changes

This PR adds a step to our app initialization, loading the translation resources associated with https://github.com/grafana/scenes/pull/1151.

### 🧪 How to test?

- All CI should pass
- The app should work as usual when running `npm run dev`

### DIFF
--- a/e2e/docker/Dockerfile.playwright
+++ b/e2e/docker/Dockerfile.playwright
@@ -1,9 +1,9 @@
-FROM mcr.microsoft.com/playwright:v1.55.1
+FROM mcr.microsoft.com/playwright:v1.56.0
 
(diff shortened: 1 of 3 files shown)


bohandley (approved):
🚀

Issue #683, opened 2025-09-08
Subpath is missing for explore in new window


mariusz-michalczyk:
Hi Team

**What happened?**
In Metrics Drilldown, buttons that open explore in new window have the sub path missing.

**How do we reproduce it?**

- Serve grafana with sub path, e.g. /my-grafana with virtual proxy
- Feed grafana with some data to have results in any metrics
- Go to metrics Drilldown
- Either click icon button on the right "Open in explore" or "Explore" in individual metrics graph menu
- Generated url doesn't contain subpath 'my-grafana' which causes page not found 404

**Environment:**
Grafana: 12.1.1
Win 10
Brrowser: Chrome  139.0.7258.157

**Datasources:**
Mimir

bohandley:
Thank you @mariusz-michalczyk and @mgyongyosi for the information. I will be opening an issue in Grafana for this and closing this issue because it is not related to Metrics Drilldown.

We have tested the subpath

```
# The full public facing url
root_url = %(protocol)s://%(domain)s:%(http_port)s/

# Serve Grafana from subpath specified in `root_url` setting. By default it is set to `false` for compatibility reasons.
serve_from_sub_path = true
```

and we have tested `useSessionStorageForRedirection = false`

Both of these are beyond Metrics Drilldown so I think to get this resolved the best course is to take this issue up in the Grafana repo.

PR #768, opened 2025-10-07
use sub url explicitly. Logs drilldown does this.

bohandley:
Fixes https://github.com/grafana/metrics-drilldown/issues/683

### ✨ Description

**Related issue(s):** https://github.com/grafana/metrics-drilldown/issues/683

Sub url is not being added when using "Oped in explore" link for some cases.

The precedent for other apps is to explicitly reference the ap sub url

https://github.com/grafana/logs-drilldown/blob/40f20334c29b5bd2ae32cb17320892c78dace4db/src/Components/ServiceScene/GoToExploreButton.tsx#L68-L72

### 📖 Summary of the changes

Use sub url explicitly in open in explore link.

### 🧪 How to test?

The build should pass.
Open in explore should always contain the sub url which can be configured here in Grafana custom.ini

https://github.com/grafana/grafana/blob/22b88988a4c8e0e2e63e6143bc0627b33ff018a0/conf/defaults.ini#L59-L63

### DIFF
--- a/src/MetricScene/PanelMenu/actions/ExploreAction.ts
+++ b/src/MetricScene/PanelMenu/actions/ExploreAction.ts
@@ -1,4 +1,5 @@
 import { type PanelMenuItem } from '@grafana/data';
+import { config } from '@grafana/runtime';
 import { getExploreURL, sceneGraph, VizPanel } from '@grafana/scenes';
 
 export class ExploreAction {
@@ -31,7 +32,9 @@ export class ExploreAction {
     return {
       text: 'Explore',
       iconClassName: 'compass',
-      onClick: () => exploreUrl?.then((url) => url && window.open(url, '_blank')),
+      onClick: () => exploreUrl?.then((url) => {
+        return url && window.open(`${config.appSubUrl}${url}`, '_blank');
+      }),
       shortcut: 'p x',
     };
   }


src/MetricScene/PanelMenu/actions/ExploreAction.ts
```diff
@@ -31,7 +32,11 @@ export class ExploreAction {
     return {
       text: 'Explore',
       iconClassName: 'compass',
-      onClick: () => exploreUrl?.then((url) => url && window.open(url, '_blank')),
+      onClick: () => exploreUrl?.then((url) => {
+        if (url) {
+          window.open(`${config.appSubUrl}${url}`, '_blank');
```
NWRichmond, line 37: Looking at some tests in `grafana/grafana` (like [this one](https://github.com/grafana/grafana/blob/84a2f41016c01aa5209346e9531354abd1aa4629/packages/grafana-data/src/utils/location.test.ts#L37)), it looks like the `appSubUrl` is expected to have a leading `/` but not a trailing `/`. If that's correct, I believe we need a `/` between the `appSubUrl` and `url`.

bohandley: @NWRichmond here's the actual leading `/` in the scenes repo from `getExploreUrl`

https://github.com/grafana/scenes/blob/badacd884623a85905be1a40655d69594fd9f27c/packages/scenes/src/utils/explore.ts#L62-L64

PR #769, opened 2025-10-07
fix: bundle size impact of i18n init

NWRichmond:
### ✨ Description

This PR is a follow-up to https://github.com/grafana/metrics-drilldown/pull/765. In that PR, I failed to detect the [enormous bundle size impact](https://github.com/grafana/metrics-drilldown/pull/765#issuecomment-3377510786) of the eager import.

### 📖 Summary of the changes

Switches to lazy imports of `@grafana/i18n` and `@grafana/scenes` utils for i18n init.

### 🧪 How to test?

- All CI should pass


bohandley:
I missed that bundle increase! Good catch!

PR #791, opened 2025-10-17
chore: reduce the number of screenshot tests

shelldandy:
As discussed in earlier elaboration sessions, the screenshot tests got a bit out of control and have started to slow us down a bit.

Lets keep only the higher level screenshot tests to keep toil to a minimum.

- Screenshots only at a high level overview per route
- Remove suffix filename `darwin/linux` that way we can just regenerate screenshots locally instad
- Make the `GroupBySelector` fixture be less flaky


NWRichmond:
Wow, so much simpler! Nicely done 👏🏼 

I'm looking forward to capturing some of the granular UI details as regular e2e test assertions. This feels like a step in the right direction.

PR #794, opened 2025-10-21
fix: playwright published action

shelldandy:
Some actions were moved to the internal folder in plugins actions here: https://github.com/grafana/plugin-ci-workflows/pull/338

Which was causing this issue:

```
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' for action 'grafana/plugin-ci-workflows/actions/plugins/package@main'.
```

[Example faulty run before](https://github.com/grafana/metrics-drilldown/actions/runs/18666279185/job/53218192429)

[Example fixed run after](https://github.com/grafana/metrics-drilldown/actions/runs/18695960320/job/53313353545)

This PR fixes that.

Also brought back the scripts to regen/delete screenshots but now you can filter just to the specific grafana version needed.

Also updated the supported grafana versions to match upstream

### DIFF
--- a/.github/workflows/playwright-published.yml
+++ b/.github/workflows/playwright-published.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - main
-      - fix/playwright-published
 
 permissions:
   contents: 'read'


.github/workflows/playwright-published.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           echo "plugin-name=$name" >> $GITHUB_OUTPUT
 
       - name: Sign and package
-        uses: grafana/plugin-ci-workflows/actions/plugins/package@main
+        uses: grafana/plugin-ci-workflows/actions/internal/plugins/package@main
```
shelldandy: this is the main change

yangkb09 (approved):
thank you miguel!!

Issue #802, opened 2025-10-23
Playwright Screenshots - Filename up to minor versions only


shelldandy:
To prevent conflicts whenever upstream bumps the dependency array in a patch version

PR #826, opened 2025-10-30
test: loosen up playwright screenshot version check

shelldandy:
Fixes: https://github.com/grafana/metrics-drilldown/issues/802

Instead of checking for full version match, lets check up to minor version match. That way even if the matrix of grafana versions changes upstream we should be better protected than having all tests fail due to no screenshot with that specific filename in them


NWRichmond (approved):
Nicely done! I think it makes a lot of sense to be agnostic to the patch version.

Issue #819, opened 2025-10-29
Metric scene panels load too slowly when there are many Loki data sources


NWRichmond:
## Problem

The eager fetching of **Related Logs** data to power the related logs count is taking precendence over the **Breakdown** and **Related Metrics** data fetching. Normally this isn't noticeable, but it becomes unbearably slow to render metrics panels in the **Breakdown** and **Related Metrics** tabs when there are many Loki data sources (hundreds or even thousands) to check for related logs.

## Solution

The solution is two-fold:
1. Let's defer any background tasks (like fetching data to determine the related logs count) until after data fetching is complete for the active tab in the Metric Scene.
2. Let's limit the number of Loki data sources that we check for related logs to something reasonable (like 25).

PR #830, opened 2025-10-31
fix: defer Loki health checks in the Metric Scene

NWRichmond:
### ✨ Description

**Related issue(s):** Supports #819

This PR is a follow-up to https://github.com/grafana/metrics-drilldown/pull/820. That PR helped to reduce latency associated with **Related logs** data fetching when on **Breakdown** or **Related metrics** tabs, by deferring the logs queries associated with providing a count of related logs.

But it didn't totally solve the problem, because Loki datasource health checks were still happening before the **Breakdown** / **Related metrics** queries. This is a problem because:
- There can be many (hundreds or even thousands) of Loki data sources to perform a health check on
- The health check can take a long time. Anecdotally, I observed a Loki data source respond as healthy over 30 seconds after waiting for the health check to complete.

### 📖 Summary of the changes

This PR solves a few problems:
1. Health checks now always happen after the panel data for the active tab (**Breakdown** / **Related metrics**) has finished loading
2. If any health check takes longer than 3 seconds, we abort the request and don't add the data source to the `healthyDataSources` list
3. We now perform health checks on a maximum of 15 data sources

**Implementation note**: Did you know that `@grafana/scenes` has a built-in [`SceneQueryController`](https://github.com/grafana/scenes/blob/bfbe80066fa1d0acafa945e77c486ad4cf618f0d/packages/scenes/src/behaviors/SceneQueryController.ts) behavior? Me neither! But it's quite handy for this use case, where we're focused on the overall data fetching picture for a particular scene.

### 🧪 How to test?

(Adapted from https://github.com/grafana/grafana/pull/98775)

I recommend using https://github.com/NWRichmond/demo-prometheus-loki-shared-labels to test this PR. If you'd like to test with other Prometheus & Loki resources, awesome!

1. Check out https://github.com/NWRichmond/demo-prometheus-loki-shared-labels and follow the Getting Started to spin up Prometheus and Loki resources on ports `9098` and `3108`, respectively
2. Create new Prometheus and Loki data sources that point to these services. The easiest way to do that is to create a `provisioning/datasources/labels-cross-reference.yaml` file containing:

```yaml
apiVersion: 1

datasources:
  - name: prometheus-labels-x-ref
    uid: prometheus-labels-x-ref
    type: prometheus
    access: proxy
    url: http://host.docker.internal:9098
    basicAuth: false
    withCredentials: false
    jsonData:
      manageAlerts: true
      prometheusType: Prometheus
      prometheusVersion: '3.2.1'
      cacheLevel: 'Low'
  - name: loki-labels-x-ref
    uid: loki-labels-x-ref
    type: loki
    access: proxy
    url: http://host.docker.internal:3108
    jsonData:
      manageAlerts: true
```

3. In Metrics Drilldown, select the new Prometheus data source (`prometheus-labels-x-ref`)
4. Select a metric and apply some labels to the filter. Some labels are shared (such as `region=us-west-2,environment=dev`) while others are not (`instance=demo-app:80`). We expect logs to only appear when the filters contain shared labels.
5. Note that when you're on the **Breakdown** or **Related metrics** tabs, the **Network** tab in DevTools shows that health check requests (`/health`) happen after metrics requests (see **Demo** below).

#### **Breakdown** tab

https://github.com/user-attachments/assets/e16496eb-441b-4c68-8308-1cc98d096060

#### **Related metrics** tab

https://github.com/user-attachments/assets/fd66b498-cae8-4c94-a117-426f1bbdd4f2


src/MetricScene/RelatedLogs/RelatedLogsOrchestrator.ts
```diff
@@ -152,13 +156,9 @@ export class RelatedLogsOrchestrator {
...
         datasource: { uid: datasource.uid },
-        queries: [],
-        key: `related_logs_check_${datasource.uid}`,
-      });
-
-      // Build and set queries
-      queryRunner.setState({
         queries: this.getLokiQueries(datasource.uid),
+        key: `related_logs_check_${datasource.uid}`,
+        $timeRange: this._metricScene.state.$timeRange,
```
NWRichmond: We weren't passing this before, but AFAIK we haven't encountered bugs related to the time ranges of related logs

src/shared/utils/utils.datasource.ts
```diff
@@ -39,7 +39,7 @@ export class DataSourceFetcher {
    * @param type - The type of data source to retrieve ('prometheus' or loki)
    * @returns Array of healthy data sources
    */
-  public async getHealthyDataSources(type: DataSourceType): Promise<DataSource[]> {
+  public async getHealthyDataSources(type: DataSourceType, limit = 25): Promise<DataSource[]> {
```
NWRichmond: I'm open to feedback about `limit = 15`. 15 is certainly a lot more than the 5 Loki data sources that we allow in the data sources drop-down for Related Logs. I think that this value should be higher than 5 to account for the potential for (a) unhealthy datasources and (b) data sources that take longer than 3s to respond as healthy, but I'm not sure how much higher.

src/MetricScene/utils/signalOnQueryComplete.ts
```diff
@@ -0,0 +1,39 @@
...
+import { type ActionViewType } from '../MetricActionBar';
+
+/**
+ * Monitors a scene's queries and publishes a completion event when all queries finish.
+ *
+ * This waits for the query lifecycle: not-running → running → complete
+ * and signals only when the transition from running to complete happens.
+ *
+ * Important: This assumes queries WILL run. Tabs without queries should
+ * signal completion explicitly rather than relying on this utility.
```
bohandley: Good detail!

src/shared/utils/utils.datasource.ts
```diff
@@ -39,7 +39,7 @@ export class DataSourceFetcher {
    * @param type - The type of data source to retrieve ('prometheus' or loki)
    * @returns Array of healthy data sources
    */
-  public async getHealthyDataSources(type: DataSourceType): Promise<DataSource[]> {
+  public async getHealthyDataSources(type: DataSourceType, limit = 15): Promise<DataSource[]> {
```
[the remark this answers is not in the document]
bohandley: You can add a tooltip via the `description` property when creating the variable! :) Similar to what we do for the main data source variable for GMD.

```
    // Set up variables for datasource selection
    const logsDataSourceVariable = new CustomVariable({
      name: VAR_LOGS_DATASOURCE,
      label: 'Logs data source',
      query: this.state.orchestrator.lokiDataSources.map((ds) => `${ds.name} : ${ds.uid}`).join(','),
      description: 'Some Loki data sources might be missing from the dropdown if they took longer than ${DS_HEALTH_CHECK_TIMEOUT_S} seconds to respond. To view logs for all Loki data sources, try using Logs Drilldown.'
    });
```

Screenshot 2025-11-04 at 4 52 18 PM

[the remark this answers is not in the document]
NWRichmond: Thank you so much @bohandley! That's so much better. Implemented in https://github.com/grafana/metrics-drilldown/pull/830/commits/f57aef4cec012bb072b8480c9e4d6ce30869b04f ✅ 

demo tooltip via description

bohandley (approved):
LGTM! Thank you for this work! 🚀

PR #839, opened 2025-11-05
fix: update query with duplicate utf-8 metric names

chenseanxy:
### ✨ Description

**Related issue(s):** None

Currently with utf-8 (otel-style) metric names, in `buildQueryExpression` queries are rendered as:
```
avg(metric.name.utf8{__ignore_usage__="", "metric.name.utf8", })
```
Which duplicates the metric name twice.

### 📖 Summary of the changes

If we're [including the name in curley braces](https://github.com/grafana/metrics-drilldown/blob/56dbac3e1538594ffb567797a4aa9ec17daaabf7/src/shared/GmdVizPanel/buildQueryExpression.ts#L39-L44) we should probably stop [prepending them from in front of the query](https://github.com/grafana/metrics-drilldown/blob/56dbac3e1538594ffb567797a4aa9ec17daaabf7/src/shared/GmdVizPanel/buildQueryExpression.ts#L52), [according to PromQL syntax](https://prometheus.io/docs/guides/utf8/#querying).

This current impl might not be ideal, but we'll have to do this workaround with the current tsqtsq usage. Suggestions welcome!

### 🧪 How to test?

Updated test cases in buildQueryExpression to reflect utf-8 syntax.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,7 +52,7 @@
         "css-loader": "6.11.0",
         "dotenv": "16.6.1",
         "eslint": "8.57.1",
-        "eslint-config-prettier": "8.10.0",
+        "eslint-config-prettier": "8.10.2",
         "eslint-import-resolver-typescript": "3.10.1",
         "eslint-plugin-deprecation": "2.0.0",
         "eslint-plugin-import": "2.32.0",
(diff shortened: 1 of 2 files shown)


chenseanxy:
Wow that was quick! Thanks, fantastic product btw!

PR #879, opened 2025-12-09
fix: handle non-error object with `message`

### LINKED ISSUE #878: Gaps in error handling when `errorObject.error` is `undefined`
## Problem

As reported [here](https://raintank-corp.slack.com/archives/C0630AYC4SY/p1765201525855939), error handling in `src/App/useCatchExceptions.ts` is currently not handling the case where `errorEvent.error` is `undefined`, here:

https://github.com/grafana/metrics-drilldown/blob/cfd64bbd5e9f80228e546af4708b23bd6db18816/src/App/useCatchExceptions.ts#L72

## Proposed solution

It seems likely that the `errorEvent` is truthy, even though the `errorEvent.error` is not. Perhaps we have an `errorEvent` like:
```js
{ message: "Some error message" }
```
, without an `error` property. In this case, we could do:

```tsx
setError(ensureErrorObject(errorEvent.error ?? errorEvent, 'Uncaught exception!'));
```
, which should create a useful error in `ensureErrorObject`.

If we're able to reproduce the undefined `errorEvent.error` case, we may refine this further.

NWRichmond:
### ✨ Description

**Related issue(s):** Fixes #878.

This PR aims to improve error handling in situations like the one described in #878, where we suspect the `errorEvent` input might look like this:

```js
{ message: "Some error message" }
```
, where `errorEvent` is not an `Error`, but rather some other object with a `message` property with a string value.

### 📖 Summary of the changes

- In `src/App/useCatchExceptions.ts`, update the `setError` call to pass `errorEvent` if `errorEvent.error` is falsey. Also add some runtime checks for types. Our previous expectations about the input to `ensureErrorObject` lead to runtime exceptions, so this PR adds some defensive coding there.
- Updates tests to reflect the updated error handling expectations.

### 🧪 How to test?

- [x] All CI should pass.

### DIFF
--- a/src/AppDataTrail/DataTrail.tsx
+++ b/src/AppDataTrail/DataTrail.tsx
@@ -56,13 +56,13 @@ import { resetYAxisSync } from '../MetricScene/Breakdown/MetricLabelsList/behavi
 import { MetricScene } from '../MetricScene/MetricScene';
 import { type PanelDataRequestPayload } from '../shared/GmdVizPanel/components/addToDashboard/addToDashboard';
 import { MetricSelectedEvent, trailDS, VAR_DATASOURCE, VAR_FILTERS } from '../shared/shared';
-import { MetricDatasourceHelper } from './MetricDatasourceHelper/MetricDatasourceHelper';
 import { reportChangeInLabelFilters, reportExploreMetrics } from '../shared/tracking/interactions';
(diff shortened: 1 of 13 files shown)


yangkb09 (approved):
LGTM - thanks for this improvement, nick!

PR #912, opened 2026-01-09
fix: bump @remix-run/router

shelldandy:
updates transitive dependency `@remix-run/router`

we're in the clear anyways since we use declarative mode, but its always better to be in the green green:

image


yangkb09 (approved):
thanks for this!

PR #926, opened 2026-01-13
chore(eslint): add @grafana/i18n rules

shelldandy:
Following these guidelines: https://grafana.com/developers/plugin-tools/how-to-guides/plugin-internationalization-grafana-11#configure-eslint-rules-for-translations

- Add @grafana/i18n ESLint plugin with two rules:
  - no-untranslated-strings: Catch untranslated strings in JSX and props
  - no-translation-top-level: Prevent t() calls at module level
- No namespace prefix for auto-generated translation keys
- Preserve all existing ESLint rules and plugin configurations

Note: some `sonarjs` disable comments were no longer needed i just ran `npm run lint:fix` and its the reflection of those warnings getting cleared.


src/AppDataTrail/DataTrail.tsx
NWRichmond: It looks like something's not quite right with the React linting:

Image

PR #873, opened 2025-12-03
feat: add exposed component for Knowledge Graph source metrics

NWRichmond:
### ✨ Description

**Related issue(s):** This PR supports #582.

### 📖 Summary of the changes

As described in https://github.com/grafana/asserts-app-plugin/pull/2403, this PR exposes a new `grafana-metricsdrilldown-app/knowledge-graph-insight-metrics/v1` component that enables a new experience in the Knowledge Graph's **RCA Workbench**. This component accepts a list of [Source Metrics](https://docs.google.com/presentation/d/1sPMtzVpCP1FahvEUUE-0-qdG710l-0ffxnjRMwQcUvA/edit?usp=sharing) provided by the Knowledge Graph, then uses the common `labels` to display all metrics related to a particular Knowledge Graph [insight](https://grafana.com/docs/grafana-cloud/knowledge-graph/introduction/insights).

This component is basically a vanilla Metrics Reducer view with:
- certain labels applied
- certain prefix filters applied, to avoid presenting `asserts:` or `ALERTS`/`ALERTS_FROM_STATE` metric panels

### Demo

Please see the **Demo** section in https://github.com/grafana/asserts-app-plugin/pull/2403.

### 🧪 How to test?

Please follow the **Testing notes** described in https://github.com/grafana/asserts-app-plugin/pull/2403.

### Notes

I'm not yet content with the Source Metrics UX. Currently, it's still not easy enough for the average user to know which of the source metrics are most worth exploring, and the **Breakdown** view tends to not be very helpful when labels contain a single series, especially when the label is already included in the label filters.

I've got some ideas about how to refine this experience in Metrics Drilldown, but would welcome any feedback / UX suggestions!

PR #958, opened 2026-01-21
feat: Knowledge Graph source metrics enhancements

### LINKED ISSUE #582: "Raw Metrics" in Asserts RCA Workbench
“Raw Metrics” in the Asserts RCA Workbench

Asserts create aggregated metrics to capture latency and etc. 

It would be valuable to be able to get back to the original metrics when in the middle of an RCA.

NWRichmond:
### ✨ Description

**Related issue(s):** Closes #582.

Building on https://github.com/grafana/metrics-drilldown/pull/873 and backed by data fetching in https://github.com/grafana/asserts-app-plugin/pull/2403, this PR adds the remaining items from the Knowledge Graph x Drilldown teams' wishlist for the MVP of the Source Metrics experience in Metrics Drilldown. The acceptance criteria are:
- When there's just a single metric in the provided `sourceMetrics` array, we should automatically select that metric instead of showing the Metrics Reducer view.
- In the **Breakdown** tab, we should filter out label names beginning with `asserts`, as they are added by the Knowledge Graph and their presence makes it harder to locate the kind of Source Metrics signals that will be interesting to users.
- When selecting a histogram metric, let's default to showing percentiles (p50, p90, and p99) instead of a heatmap.
- When no `sourceMetrics` are provided, we continue to fall back to selecting the relevant Asserts recording rule metric.

### 📖 Summary of the changes

- In `src/exposedComponents/SourceMetrics/SourceMetrics.tsx`, enables automatic selection of a source metric when the `sourceMetrics` array only contains one item. Introducing another scenario with its own treatment motivated a small refactor to make the different logical branches cleaner/clearer.
- Added a `FilterGroupByAssertsLabelsBehavior` behavior to handle the filtering of label names starting with `asserts` in the Breakdown view. This is complemented by a new `GroupByOptionsLoadedEvent` in `src/MetricScene/Breakdown/GroupByVariable.tsx`.
- Added a `HistogramPercentilesDefaultBehavior` behavior to handle the selection of the percentiles view (instead of the default heatmap view) by default.
- Removes the **Give feedback** button in Embedded Mode, as it could be confusing to the user, who could reasonably expect that we're soliciting feedback in the parent experience rather than in the exposed component from Metrics Drilldown.

### 🧪 How to test?

For general setup of Metrics Drilldown in the RCA Workbench, follow the **Testing notes** described in https://github.com/grafana/asserts-app-plugin/pull/2403, with the exception that we should use the `main` branch of `asserts-app-plugin` and the `582/knowledge-graph-source-metrics-enhancements` branch of `metrics-drilldown`.

From there, you could follow the actions shown in the **Demo** below by visiting http://localhost:3000/a/grafana-asserts-app/assertions?start=now-6h&end=now&search=Show%20all%20Services&view=BY_ENTITY.

### Demo

When opening an Insight's metrics, a source metric (distinguishable because it's not an `asserts:`-prefixed recording rule) is selected. Its **Breakdown** tab doesn't include label names starting with `asserts`:

https://github.com/user-attachments/assets/6a98f206-c444-4b8c-9b00-d547e6f0a53d

When investigating metrics for an insight like a **p99_breach** that has a histogram-type source metric, we display that histogram metric using percentiles instead of defaulting to a heatmap, so users can quickly compare the p99 with other useful percentiles: 

https://github.com/user-attachments/assets/4f2b3684-fb2a-470e-8720-f503626568d5

### DIFF
--- a/src/exposedComponents/SourceMetrics/SourceMetrics.tsx
+++ b/src/exposedComponents/SourceMetrics/SourceMetrics.tsx
@@ -42,7 +42,7 @@ function extractUniquePrefixes(metricNames: string[]): string[] {
 
 type SourceMetricsScenario = 'single_source_metric' | 'multiple_source_metrics' | 'recording_rule_fallback' | 'missing_metric_information';
 
-function getSourceMetricsScenario(sourceMetrics?: Array<{ metricName: string; labels: PromQLLabelMatcher[] }>, parsedPromQLQuery?: ParsedPromQLQuery): SourceMetricsScenario {
+function getSourceMetricsScenario(sourceMetrics: Array<{ metricName: string; labels: PromQLLabelMatcher[] }> | undefined, parsedPromQLQuery: ParsedPromQLQuery | undefined): SourceMetricsScenario {
   if ((!sourceMetrics || sourceMetrics.length === 0) && parsedPromQLQuery) {
     return 'recording_rule_fallback';
   }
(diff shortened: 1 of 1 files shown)


src/exposedComponents/SourceMetrics/SourceMetrics.tsx
```diff
@@ -72,47 +95,63 @@ const KnowledgeGraphSourceMetrics = ({
...
-
-  // Set URL params for prefix filters BEFORE creating the trail
-  // This leverages Scenes' built-in URL sync in MetricsFilterSection
-  if (sourceMetricPrefixes?.length) {
-    const prefixUrlParam = `${embeddedTrailNamespace}-${metricFilters.prefix}`;
-    locationService.partial({ [prefixUrlParam]: sourceMetricPrefixes.join(',') }, true);
-
-    const sortByUrlParam = `${embeddedTrailNamespace}-var-${VAR_WINGMAN_SORT_BY}`;
-    locationService.partial({ [sortByUrlParam]: 'alphabetical' }, true);
+  const scenario = getSourceMetricsScenario(sourceMetrics);
```
bohandley: Do we want to pass in `parsedPromQLQuery` here?

Issue #934, opened 2026-01-16
[1.1.1] Radio button inputs in Prometheus function config missing label association


yangkb09:
# [1.1.1] Radio button inputs in Prometheus function config missing label association

**Parent Issue**: #930 

## Issue Description
Radio button inputs in the "Configure the Prometheus function" modal lack proper label association. This pattern repeats for each function option (Average, Sum, Standard deviation, Percentiles, Minimum and maximum).

## Location
- **View**: Prometheus function configuration modal
- **Element**: `<input type="radio" name="select-config" tabindex="0" checked="">` (pattern repeats for each function option)

## Current Behavior
Screen readers cannot announce which Prometheus function option each radio button represents.

## Expected Behavior
Screen reader should announce the function name (e.g., "Average (default)", "Sum", "Standard deviation") when each radio button receives focus.

## axe DevTools Output
- **Rule**: label (Form elements must have labels)
- **Impact**: Critical
- **Element**: `<input type="radio" name="select-config" tabindex="0">`

*Note: This violation appears multiple times (once per function option in the modal).*

## Suggested Fix
*Note: Suggestions by Claude. Confirm validity before implementation.*

- Add `aria-label` to each radio input with the function name
- Or wrap each radio input in a `<label>` element with visible text
- Or use `aria-labelledby` to reference the visible function name text

*Note: This is a repeating pattern. The fix should address all function options in the modal (Average, Sum, Standard deviation, Percentiles, Minimum and maximum).*

## Acceptance Criteria
- [ ] Element has proper accessible name/label
- [ ] axe DevTools reports no violations
- [ ] VoiceOver announces meaningful text alternative

PR #996, opened 2026-01-29
fix(a11y): add aria-label to Prometheus function config radio inputs

yangkb09:
### ✨ Description

**Fixes:** https://github.com/grafana/metrics-drilldown/issues/934  

Adds an accessible name to the radio inputs in the "Configure the Prometheus function" modal so screen readers can announce which option each radio represents (e.g. "Average (default)", "Sum", "Standard deviation").

Pre-changes

https://github.com/user-attachments/assets/b0ddc779-b5a5-44fc-bf89-a657c7c8e157

Post-changes

https://github.com/user-attachments/assets/54efbe3e-acec-46b7-85c4-03dc828342b1

### 📖 Summary of the changes

- In `WithConfigPanelOptions`, the radio input now has an `aria-label` derived from the preset’s display name.
- The label comes from `body.state.panelConfig.title` (the same title used for the panel), with a fallback to `presetId` if the title is missing.
- No new props or changes to `ConfigurePanelForm`; only the single component that renders the radios was updated.

### 🧪 How to test?

1. Open a metric’s "Configure" flow so the Prometheus function config modal is visible (grid of options with a radio per row).
2. Run axe DevTools on the modal and confirm there are no "Form elements must have labels" violations on the radio inputs OR
3. With a screen reader (e.g. VoiceOver), focus each radio and confirm it announces the function name (e.g. "Average (default)", "Sum") instead of an unlabeled control.


bohandley (approved):
LGTM! Thank you for this work!

PR #1030, opened 2026-02-04
fix(ci): use pull_request_target to allow labeling fork PRs

nicwestvold:
### ✨ Description

PR #955 (a fork contribution) revealed that the PR title validation workflow fails for fork PRs because the `pull_request` event lacks write permissions to add labels.

### 📖 Summary of the changes

Changed the workflow trigger from `pull_request` to `pull_request_target`, which runs in the context of the base repository and has write permissions to add labels.

**Why this is safe:** The workflow only reads PR metadata (the title) to validate conventional commit format and add a label. It never checks out or executes code from the fork, so there's no risk of running untrusted code with elevated permissions.

### 🧪 How to test?

1. Merge this PR
2. Re-run the PR title validation workflow on #955 
3. Verify the workflow passes and the appropriate label is added

### DIFF
--- a/.github/workflows/pr-title-validation.yml
+++ b/.github/workflows/pr-title-validation.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: PR Conventional Commit Validation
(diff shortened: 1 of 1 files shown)


bohandley (approved):
Looks good to me, and thank you for checking with security to audit this! 🚀

PR #1113, opened 2026-03-02
fix: prevent native histogram double-rebuild race condition

nicwestvold:
### ✨ Description

Native histogram metrics could fail to load: `subscribeToStateChanges` set `panelConfig.type = 'heatmap'` but omitted `metricType: 'native-histogram'`, so queries never updated. A secondary race allowed both detection paths to independently trigger `buildVizPanel()`.

### 📖 Changes

- Set `metricType: 'native-histogram'` in the `subscribeToStateChanges` path
- Fix stale closure: read `this.state.panelConfig` at resolution time
- Skip state update if the other path already applied heatmap type
- Unsubscribe on first `Done` event unconditionally
- Replace `!` assertions in query builders with guards that warn and skip

### 🧪 How to test?

- Open a native histogram metric; verify it renders as a heatmap with native histogram queries
- Automated: new unit tests in `GmdVizPanel.test.ts` and the three query builder test files


bohandley:
My thoughts ([also shared on slack](https://raintank-corp.slack.com/archives/C0630AYC4SY/p1776341763529249))

It's a Scenes pattern + Prometheus API issue, so not necessarily a native histogram thing.

Whenever we build a panel, we need the parts. The parts are Prometheus-isms from filters to metric to type, etc. The flaky thing is metric type. We either get this from the naming convention or meta data. Because native histograms are the newest, they have the worst inconsistencies for metadata. They also break our naming convention.

I see that in the PR we are judging the dataframe type as heatmap cells, which I think is ultimately better than trusting the metadata or the naming convention.

I wonder if we don't just default to this data frame check instead of doing the metadata check at all since it is so flaky? I'll make an issue for this 😄

src/shared/GmdVizPanel/GmdVizPanel.tsx
```diff
@@ -202,18 +209,29 @@ export class GmdVizPanel extends SceneObjectBase<GmdVizPanelState> {
...
+        // Always unsubscribe on first Done event — prevents multiple firings.
+        bodySub.unsubscribe();
+
         const dataFrameType = newState.data.series?.[0]?.meta?.type;
         if (!dataFrameType) {
           return;
         }
 
         if (dataFrameType === DataFrameType.HeatmapCells) {
+          // Guard: if checkMetricMetadata (Path A) already rebuilt to heatmap, skip.
```
bohandley, line 221: This is always consistent

bohandley (approved):

PR #1173, opened 2026-03-31
feat: add Knowledge Graph annotation overlays

rdubrock:
Closes https://github.com/grafana/asserts-app-plugin/issues/2920

https://github.com/user-attachments/assets/05ec1ad8-f8d9-4f0c-a529-a8f6e27ebb00

Screenshot 2026-04-02 at 11 11 07 AM
Screenshot 2026-04-01 at 5 58 38 PM

## Summary
- Adds Knowledge Graph health-state annotations (critical, warning, info) as overlays on timeseries and percentile panels
- Gated behind the `kgAnnotationsInMetricsDrilldown` feature toggle and requires the KG datasource to be provisioned
- Includes an "Insights" toggle control in the header and a reactive behavior that updates annotation layers when filters or datasource change
- Enables multi-lane annotation display on all timeseries/percentile panels

## Test plan
- [ ] Verify annotations do not appear when feature toggle is disabled
- [ ] Verify annotations do not appear when KG datasource is not provisioned
- [ ] Enable toggle and confirm critical/warning/info annotations render on timeseries panels
- [ ] Toggle "Insights" switch off and confirm annotations hide
- [ ] Change filters and verify annotation layers update reactively
- [ ] Confirm no regressions on percentile panels

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/shared/featureFlags/openFeature.ts
+++ b/src/shared/featureFlags/openFeature.ts
@@ -11,6 +11,11 @@ import { logger } from '../logger/logger';
  * See {@link https://github.com/grafana/deployment_tools/blob/master/ksonnet/environments/hosted-grafana/waves/feature-toggles/goff/drilldown/metrics/flag-definitions.libsonnet | Metrics Drilldown GoFF flag definitions} for the source of truth.
  */
 const goffFeatureFlags = {
+  kgAnnotationsInMetricsDrilldown: {
+    valueType: 'boolean',
+    values: [true, false] as const,
+    defaultValue: false,
+  },
   'drilldown.metrics.default_open_sidebar': {
     valueType: 'string',
     values: [
@@ -41,7 +46,7 @@ const goffFeatureFlags = {
     defaultValue: 'excluded',
     trackingKey: 'experiment_grafana_assistant_quick_search_tab_test',
   },
-} as const satisfies Record<`drilldown.metrics.${string}`, FeatureFlag>;
+} as const satisfies Record<string, FeatureFlag>;
 
 /**
  * This discriminated union captures the different types of feature flags that can be evaluated.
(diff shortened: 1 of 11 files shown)


bohandley (commented):
Love it!

| Metrics Reducer | Metric Scene |
| :--- | :---: |
| Image | Image |

I like all of this:
- Dual-path design is clean: direct entity layers for `EntityMetrics`, behavior-reactive layers for the main trail -- right call for both contexts.
- `KgAnnotationBehavior` subscribing to `VAR_FILTERS` and `VAR_DATASOURCE` via `subscribeToState` is idiomatic Scenes.
- `lookupKey` deduplication in the behavior avoids redundant layer rebuilds when filters haven't meaningfully changed.
- `isKgAnnotationsAvailable()` correctly gates on both the feature toggle and datasource presence before doing any work.
- `fromLabels` query mode is smart -- lets the KG backend resolve entity identity from metric label context at query time, no explicit entity required.
- i18n keys added for all user-visible strings.
- `multiLane: true` applied consistently across timeseries and percentile panel builders.
- Test coverage hits the meaningful paths (flag off, datasource missing, both present).

---

Just one concern with the UX where annotations appear on every panel. The current implementation adds annotation overlays to all timeseries and percentile panels in the metrics reducer and labels breakdown. Showing insight annotations in the metrics scene (main trail) makes sense, but blanketing every panel in the labels breakdown is likely too much. This should go to UX or the squad for a sanity check on where annotations belong and where they add noise.

nicwestvold (approved):

src/shared/featureFlags/openFeature.ts
```diff
@@ -11,6 +11,11 @@ import { logger } from '../logger/logger';
  * See {@link https://github.com/grafana/deployment_tools/blob/master/ksonnet/environments/hosted-grafana/waves/feature-toggles/goff/drilldown/metrics/flag-definitions.libsonnet | Metrics Drilldown GoFF flag definitions} for the source of truth.
  */
 const goffFeatureFlags = {
+  kgAnnotationsInMetricsDrilldown: {
```
bohandley, line 14: Example of Open feature toggle usage

PR #1207, opened 2026-04-22
fix(native-histogram): rely solely on data frames for native histogram detection

### LINKED ISSUE #1199: Native histogram detection should rely on data frames, not metadata
## Background

The plugin detects metric types using a layered strategy:

1. **Name (sync, instant)** - naming conventions: `_total`/`_count`/`_sum` for counters, `_bucket` for classic histograms, `_up` for status-updown, etc.
2. **Metadata (async)** - fetches Prometheus metadata to correct gauge/counter misidentification and to detect native histograms
3. **Data frame (async, after first query)** - inspects the returned data frame type; `HeatmapCells` indicates a native histogram

This hierarchy works for every metric type except native histograms, because native histograms are the only type with no naming convention. They are indistinguishable from gauges by name, by design. So layer 1 always falls through to gauge. Layer 2 (metadata) is then the first real signal.

## The problem with metadata for native histograms

Native histogram support in Prometheus is relatively new and the `type: histogram` metadata field is not consistently populated across Prometheus versions and exporters. This is an ecosystem limitation, not a bug in this plugin. It means layer 2 is unreliable for native histograms specifically, even though it works fine for gauge/counter correction (issue #698).

Because metadata is unreliable, a data frame check (layer 3) was added as a fallback: if the first query returns `DataFrameType.HeatmapCells`, the panel switches to heatmap. This is more reliable because the data itself is ground truth - if Prometheus sends HeatmapCells, the metric is a native histogram regardless of what the name or metadata says.

## The current design is creating bugs

Two independent detection paths now run in parallel for native histograms:

- **Path A** (`checkMetricMetadata`): fetches metadata, sets `metricType` and `panelConfig.type`
- **Path B** (`subscribeToStateChanges`): waits for first query result, checks data frame type

These paths race against each other. #1113 (merged) fixed a bug where Path B set `panelConfig.type = 'heatmap'` without also setting `metricType = 'native-histogram'`, causing the panel to render as heatmap but run gauge queries. See that PR for the specific changes. Beyond the missing field, the PR also had to add guards to manage the interaction between the two paths, including stale closure fixes and double-setState prevention. That added complexity is a symptom of the underlying design problem: two overlapping detection paths for the same thing.

## Proposal

Remove native histogram detection from `checkMetricMetadata`. Specifically, remove the block:

```js
if (metricTypeFromMetadata === 'native-histogram') {
  stateUpdate.metricType = 'native-histogram';
  panelConfigUpdate.description = ...;
  if (!discardPanelTypeUpdates) {
    panelConfigUpdate.type = 'heatmap';
  }
}
```

The gauge/counter correction logic in `checkMetricMetadata` (issue #698) is unrelated and stays untouched. `checkMetricMetadata` continues to exist and run for that purpose.

Path B (data frame detection) becomes the sole method for native histogram detection. This is the correct approach because data frames are the only reliable signal for this metric type.

## Trade-off

Path B requires the first query to complete before the panel switches to heatmap. There will be a brief initial render as a timeseries panel. This is acceptable given the reliability benefit and the fact that Path A was already unreliable for a significant portion of users.

## Future state

When native histogram metadata becomes consistently populated and enforced across Prometheus versions, a clean early-detection path can be added - not as a parallel fallback, but as a deliberate fast path that skips the first query wait when metadata is conclusive. That is a separate piece of work and should not be done until the metadata situation is stable.

yangkb09:
### ✨ Description

Fixes: https://github.com/grafana/metrics-drilldown/issues/1199

Remove native histogram detection from the metadata path so that data frame inspection is the sole detection method.

### 📖 Summary of the changes

Native histograms have no naming convention and are indistinguishable from gauges by name. Detection previously ran via two competing paths:

  - Path A (`checkMetricMetadata`): async Prometheus metadata fetch. Unreliable because native histogram metadata is not consistently populated across Prometheus versions and exporters, causing race conditions and incorrect panel types.
  - Path B (`subscribeToStateChanges`): data frame inspection after the first query completes. Reliable ground truth — if Prometheus returns `HeatmapCells` frames, the metric is definitively a native histogram.

This PR removes Path A's native histogram handling entirely. `checkMetricMetadata` now only corrects gauge/counter misidentification (issue #698). Path B remains unchanged as the sole native histogram detection method.

Trade-off: the panel renders briefly as a timeseries before switching to heatmap on the first query completion. This is acceptable given the reliability benefit.

### 🧪 How to test?

  1. Open a Prometheus datasource that exposes native histogram metrics.
  2. Select a native histogram metric in Metrics Drilldown.
  3. Verify the panel renders as a heatmap.
  4. Verify gauge and counter metrics still display correctly (no regression in gauge/counter type correction).
  5. Run unit tests: pnpm jest GmdVizPanel.test


bohandley (commented):
Thank you for this work @yangkb09! It's looking cleaner and easier to reason about.

There are a few issues due to the complexity of native histograms and the waterfall of things that happen in metrics drilldown.

1. If a user has changed function configuration, the check does not occur. This is spoken in more detail from copilot's review: 

> Removing native-histogram handling from checkMetricMetadata means native histograms are now only detected in subscribeToStateChanges. However that subscription is disabled when discardPanelTypeUpdates=true (which happens when panelOptions.type or a saved preference sets panel type). In that case a native-histogram metric will never get metricType='native-histogram', so histogram-specific panel types (e.g. a saved 'percentiles' preset) will build non-histogram queries indefinitely. Consider allowing the data-frame inspection path to still update metricType (and description) even when panel type updates are discarded, and only gate the panelConfig.type switch on discardPanelTypeUpdates (or add a separate frame-based probe query when starting in non-timeseries panel types).

2. There is a regression with low risk, also identified by copilot:

> checkMetricMetadata captures {metric, metricType} before awaiting getMetricType(). If state.metricType changes while awaiting (e.g. the data-frame path sets metricType to 'native-histogram' after the first query), this method can still apply gauge/counter corrections based on the stale pre-await metricType and overwrite the newer metricType. Re-check this.state.metric/metricType after the await (or compare to the captured values and bail) before calling setState.

bohandley (approved):
Looks good! In testing this, I found a bug with the function configuration. It does not correctly update when native histograms are identified, but I think this existed before this PR.

I created an issue if you would like to follow up on this histogram related bug @yangkb09 
https://github.com/grafana/metrics-drilldown/issues/1228

Otherwise, this works for the panels! Thank you! 🚀

PR #1209, opened 2026-04-23
chore(webpack): fix Immutable.js build warnings

shelldandy:
## Summary
- Adds a webpack resolve alias to map `immutable` to its CJS build (`dist/immutable.js`)
- `@react-awesome-query-builder/core@6.6.15` imports `immutable` with a default import, but immutable v5.x only provides named exports in the ESM build — the CJS build allows webpack to synthesize a default export
- Eliminates 60 build warnings

## Test plan
- [x] `pnpm build` produces no immutable-related warnings (only asset size warning remains)
- [x] `pnpm dev` — verify the plugin loads correctly in Grafana and pages using the query editor still work


bohandley (approved):
LGTM!

PR #1211, opened 2026-04-24
chore: add mise.toml and align CI node version

nicwestvold:
### ✨ Description

Add `mise.toml` for local dev tooling and align CI Node version references.

### 📖 Summary of the changes

- Add `mise.toml` with Node 24.15.0 and corepack enabled
- Update `is-compatible.yml` and `bundle-stats-create.yml` to read Node version from `package.json` engines instead of hardcoding `22`, matching the existing pattern in `bundle-stats-compare.yml`

### 🧪 How to test?

- [ ] Verify CI workflows pass on this PR
- [ ] Run `mise install` locally and confirm correct Node version

### DIFF
--- a/.github/workflows/bundle-stats-create.yml
+++ b/.github/workflows/bundle-stats-create.yml
@@ -21,13 +21,19 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Get Node.js version from package.json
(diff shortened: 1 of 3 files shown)


shelldandy (approved):
that looks great thanks

PR #1225, opened 2026-04-28
fix: type InlineBanner error prop as unknown to prevent crash

shelldandy:
## Summary
- Changes `InlineBanner`'s `error` prop type from `Error` to `unknown`, aligning the type with its internal `ensureErrorObject` implementation which already handles arbitrary types defensively
- Replaces unsafe `data.errors![0] as Error` with `data.errors?.[0]` in `MetricLabelValuesList` — this was the crash reported by Josh Hunt when adding a label filter on dev (React 19 + Firefox)
- Applies the same optional chaining cleanup in `QueryResultsScene`

## Test plan
- [x] `tsc --noEmit` passes
- [x] All 531 tests pass
- [ ] Verify on dev with React 19 feature flag that adding a label filter no longer crashes
- [ ] Verify error banners still render correctly when queries actually fail

### DIFF
--- a/src/App/InlineBanner.tsx
+++ b/src/App/InlineBanner.tsx
@@ -9,7 +9,7 @@ type InlineBannerProps = {
   severity: AlertVariant;
   title: string;
   message?: string | React.ReactNode;
-  error?: Error;
+  error?: unknown;
   errorContext?: ErrorContext;
   children?: React.ReactNode;
 };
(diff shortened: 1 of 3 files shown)


nicwestvold (approved):

bohandley (approved):
Love it! Thank you 🙏🏻 🚀

PR #1227, opened 2026-04-28
chore: link upstream Scenes issue for expressionBuilder bypass

shelldandy:
## Summary

- Filed upstream issue [grafana/scenes#1436](https://github.com/grafana/scenes/issues/1436) for `DrilldownDependenciesManager.getFilters()` ignoring `expressionBuilder` on `AdHocFiltersVariable`
- Updated the TODO in `patchSceneQueryRunner.ts` with the actual issue URL

## Context

Research into the feasibility of downgrading `@grafana/scenes` from v7 to v6 revealed:

- **Downgrading is low complexity** — all 48 APIs we import exist in v6, peer deps are compatible with our v13 stack, no deep imports affected
- **It would hide the `__name__` symptom** — v6 uses a weaker global variable lookup that doesn't find our `AdHocFiltersVariable`, so `request.filters` stays empty
- **But the underlying bug exists in v6 too** — `getFilters()` ignores `expressionBuilder` in both versions; v7's hierarchy walk just exposed it
- **Recommendation**: keep v7 + our monkey-patch and push for the upstream fix

## Test plan

- [ ] `pnpm typecheck` passes
- [ ] `pnpm test -- --testPathPattern=patchSceneQueryRunner` passes


bohandley:
Here is the patch for this regression: https://github.com/grafana/metrics-drilldown/pull/1226

PR #1244, opened 2026-05-11
refactor(error-handling): replace global error capture with scoped React Error Boundary

### LINKED ISSUE #1224: Spike: Research using ErrorBoundary component over current error handling strategy
### Problem

Our error capturing requires updating a list of errors to ignore (fragile) and often causes fatal crashes as a result of capturing all the errors.

- Different browsers emit different error messages, so trying to match on that seems fragile
- listening to the error event from window is overzealous

### Solution

Use ErrorBoundary component from `grafana/ui` which reports errors to faro.

### Research questions to help move this along

1. Will ErrorBoundary ever cause fatal crashes?
2. If we remove our custom error capturing solution, will we need to change our error dashboard in FE O11y?
3. Are there any edge cases we are missing if we switch to ErrorBoundary component?

shelldandy:
### ✨ Description

**Related issue(s):** https://github.com/grafana/metrics-drilldown/issues/1224
Resolves https://github.com/grafana/metrics-drilldown/issues/1224

Replace the fragile `errorsToIgnore` exact-string allowlist and global `window.error`/`unhandledrejection` listeners with a properly scoped `<AppErrorBoundary>`, Faro-native regex `ignoreErrors`, and explicit `ChunkLoadError` recovery. This is Option C from the [bet document](https://github.com/grafana/metrics-drilldown/blob/chore/error-boundary/BET-error-capture.md).

### 📖 Summary of the changes

**Step 1 — Regex ignore patterns** (`faro.ts`)
- Replaced 5 exact-string `errorsToIgnore` + `shouldIgnoreError()` with 7 regex `ignoreErrorPatterns` covering ResizeObserver, extension URLs (`chrome-extension://`, `moz-extension://`), Safari extension prefix, and cancelled promise rejections.
- Passed directly to Faro's native `ignoreErrors` config — single source of truth.

**Step 2 — Scoped `<AppErrorBoundary>`** (`AppErrorBoundary.tsx`)
- Thin wrapper over `@grafana/ui`'s `ErrorBoundary` (render-prop API).
- Custom `errorLogger` routes errors through our **isolated** Faro instance via `logger.error()`, preserving the `handheldBy: 'React error boundary'` taxonomy for FE O11y dashboards. The default `@grafana/ui` logger would write to the global Faro and we'd lose `grafana-metricsdrilldown-app-*` events.
- `dependencies={[pathname]}` enables navigation-based auto-recovery (no hard reload needed).
- Wraps providers inside the router (ErrorView needs `useNavigate`/`useLocation`).

**Step 3 — Removed global listeners** (`useCatchExceptions.ts`)
- Replaced the entire hook with a no-op shim: `return [undefined, () => {}]`.
- Removed `useCatchExceptions` imports and dead `ErrorView` error branches from all 4 exposed components (`EntityMetrics`, `LabelBreakdown`, `MiniBreakdown`, `SourceMetrics`). Since the shim always returns `[undefined, ...]`, their `error ? <ErrorView /> : <Content />` ternary was unreachable dead code.
- `SourceMetrics` retains its own `ErrorView` import for the `missing_metric_information` scenario (line 167), which is unrelated to the global error hook.
- `ensureErrorObject()` extracted to `errorUtils.ts` for `InlineBanner`.
- Faro's built-in `errors` instrumentation continues to capture `window.onerror`/`onunhandledrejection` for telemetry — no data loss.

**Step 4 — ChunkLoadError recovery** (`AppErrorBoundary.tsx`)
- Detects chunk errors by `error.name === 'ChunkLoadError'` or message patterns (`/Loading chunk .* failed/`, `/Failed to fetch dynamically imported module/`).
- First failure: auto-reloads after 2s with a "reloading…" message.
- Second failure (sessionStorage guard): shows manual "Reload now" button — prevents infinite reload loops.
- This is the gate that prevents repeating the [PR #233](https://github.com/grafana/metrics-drilldown/pull/233) regression.

**Step 5 — Cleanup**
- Deleted `useCatchExceptions.test.ts` (261 lines). Coverage replaced by `AppErrorBoundary.test.tsx` (13 tests) and `errorUtils.test.ts` (4 tests).
- Updated `InlineBanner.tsx` import path.
- Extracted `chunk-load-recovery` i18n keys to locale file.

### 🧪 How to test?

**Automated:**
- `npx jest --no-coverage --forceExit` — 537 tests pass (57 suites)
- `npx tsc --noEmit` — clean

**Manual verification (pre-merge):**
- [ ] Start dev server, navigate normally — no ErrorView shown
- [ ] Trigger a render error (e.g. temporarily `throw` in a component) — ErrorView appears with "Fatal error!" and stack trace
- [ ] Navigate away from the broken route — boundary auto-recovers
- [ ] Rename a built chunk file to simulate 404 → "reloading…" banner appears, page auto-reloads
- [ ] If chunk stays broken after reload → "Reload now" button appears (no infinite loop)
- [ ] Install a browser extension that throws errors → MD continues to render, no "Fatal error!" screen
- [ ] Check FE O11y dashboard: errors still appear with `handheldBy: 'React error boundary'` and correct `app.name`

**Follow-ups (out of scope):**
- Exposed components need their own error boundaries (they currently have no error handling after the dead-code removal)
- `MetricSelectScene` cancelled promise handling refactor
- Share ChunkLoadError pattern with logs-drilldown ([#1713](https://github.com/grafana/logs-drilldown/issues/1713))
- `docs/error-handling.md` documentation


BET-error-capture.md
```diff
@@ -0,0 +1,203 @@
...
+- **Rollback plan:** Single‑PR change. If a regression surfaces in dev/ops, revert the PR; the previous `useCatchExceptions` + `errorsToIgnore` exact list returns. No data or config migrations are involved.
+
+## Out of scope
+
+- Replacing or refactoring the existing `<ErrorView>` UI itself. (We reuse it as the boundary's fallback.)
+- Reworking the `MetricSelectScene` cancellation handling. Tracked as a follow‑up.
+- Changing the global Faro initialization timing or the `beforeSend` URL filter. (Both are working.)
+- Per‑panel boundaries inside `<Trail>`. The bet adds an app‑level boundary; finer‑grained scenes/breakdown boundaries (the model `sql-drilldown`'s [`RowsScene.tsx`](https://github.com/grafana/sql-drilldown/blob/main/src/pages/Home/scenes/RowsScene.tsx) and `logs-drilldown` [PR #1735](https://github.com/grafana/logs-drilldown/pull/1735) follow) are a sensible follow‑up if signal volume warrants.
+- Cross‑drilldown harmonization of the logger/Faro layer. `logs-drilldown` and `traces-drilldown` use `@grafana/runtime`'s `logError` against the global Faro; MD and `profiles-drilldown` use isolated Faro instances. This split is a separate platform‑level conversation.
```
shelldandy: just make sure to remove this file after pr review is done

src/App/useCatchExceptions.ts
```diff
@@ -1,123 +1,12 @@
...
-      colno: errorEvent.colno?.toString(),
-    });
-    return false;
-  }
-
-  // If it's not one of the special cases above, it's probably a legitimate application error
-  return true;
-}
-
 export function useCatchExceptions(): [Error | undefined, React.Dispatch<React.SetStateAction<Error | undefined>>] {
```
bohandley, line 8: Can we just remove this entirely? I cannot find anywhere it is used 🤔

src/App/App.tsx
```diff
@@ -22,29 +21,22 @@ const prometheusDatasources = Object.values(config.datasources).filter(isPrometh
...
-    return (
-      <div className={styles.appContainer} data-testid="metrics-drilldown-app">
-        <ErrorView error={error} />
-      </div>
-    );
-  }
-
   if (!prometheusDatasources.length) {
     return <Onboarding />;
   }
```
bohandley, line 29: Just flagging a few things our new strategy is missing. Non blocking but can we create an issue to look into handling errors for these?

bohandley (approved):
Overall +1. The shape matches what the BET doc lays out, and the four-step structure (regex ignoreErrors, app-scoped boundary, drop window listeners, chunk recovery) checks out in the diff.

Boundary placement and recovery look right. Custom errorLogger at AppErrorBoundary.tsx:21-25 preserves handheldBy: 'React error boundary', addressing uncertainty #1 in the BET doc. Navigation recovery via dependencies={[pathname]} is wired correctly per AppErrorBoundary.tsx:77 and covered by the "recovers when pathname changes" case in AppErrorBoundary.test.tsx. I exercised the Fatal path itself with a render throw and the ErrorView renders as expected.

One flag for us in the future, not a blocker: the boundary lives at the app root (App.tsx:33), so a TypeError in any one scene takes the whole plugin to the Fatal UI. The BET doc calls this out as uncertainty #5 and references sql-drilldown `ErrorBoundaryAlert` precedent. I'd be in favor of a route-level boundary as a follow-up so a broken QueryResultsScene doesn't take down the metrics list. Not for this PR but in a follow and would be great to make an issue 🥳 

Great work, thank you @shelldandy!

PR #1256, opened 2026-05-18
fix: strip drilldown legend from Explore links

### LINKED ISSUE #1248: bug: hardcoded legendFormat carried into Explore causes duplicate series names
### What happens

Opening a metric panel in Explore from Drilldown carries over a hardcoded legend. The series end up looking wrong or confusing (e.g. multiple series sharing the same name after a query edit), and the legend setting is collapsed by default in Explore so he user has no clear way to
understand why or fix it.

### Where the legendFormat is set

- `src/shared/GmdVizPanel/types/timeseries/getTimeseriesQueryRunnerParams.ts`
- `src/shared/GmdVizPanel/types/stat/getStatQueryRunnerParams.ts`
- `src/shared/GmdVizPanel/types/percentiles/getPercentilesQueryRunnerParams.ts`
- `src/shared/GmdVizPanel/types/statushistory/getStatushistoryQueryRunnerParams.ts`

### Where the link is built

`src/MetricScene/PanelMenu/actions/ExploreAction.ts` calls
`getExploreURL(panelData, ..., (query) => ...)`. The current transform
callback only strips `__ignore_usage__` from `expr`.

### Proposed fix

Strip `legendFormat` in the `ExploreAction` transform callback so the Explore
URL does not carry Drilldown's hardcoded legend, while leaving Drilldown's
own panels untouched.

### Repro

1. Open a metric in Drilldown (no group-by).
2. Panel menu → Explore.
3. In Explore, edit the query to add `by (some_label)`.
4. Observe all series in the legend share the same name (e.g. `avg`).
5. Expand `Options → Legend` and clear the field — series names now render
   correctly from labels.

omribz156:
### ✨ Description

**Related issue(s):** https://github.com/grafana/metrics-drilldown/issues/1248
Resolves https://github.com/grafana/metrics-drilldown/issues/1248

Opening a Metrics Drilldown panel in Explore should not carry over Drilldown's hardcoded legend format, because editing the query in Explore can then leave multiple series with the same fixed legend.

### 📖 Summary of the changes

- Extracted the Explore query transform used by `ExploreAction` so it can be tested directly.
- Removed `legendFormat` only from queries marked with `fromExploreMetrics: true`, which are the Drilldown-generated queries that set hardcoded legends.
- Kept the existing `__ignore_usage__` label cleanup behavior unchanged.
- Added unit coverage for removing Drilldown legends while preserving custom legends on unmarked queries.

This was implemented with Codex assistance, with the patch kept focused and manually reviewed before sending.

### 🧪 How to test?

- `pnpm exec jest src/MetricScene/PanelMenu/actions/ExploreAction.test.ts --runInBand --no-coverage --detectOpenHandles`
- `pnpm run typecheck`
- `pnpm exec eslint src/MetricScene/PanelMenu/actions/ExploreAction.ts src/MetricScene/PanelMenu/actions/ExploreAction.test.ts`
- `git diff --check`


shelldandy (approved):
lgtm!

bohandley (approved):
LGTM! Thank you for your contribution @omribz156

bohandley:
We're close @omribz156. Just need signatures on the commits 

`Commits must have verified signatures.`

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

shelldandy:
@omribz156 could you amend your commits to include a gpg signature.
We need verified commits as part of our security protocols. 

I think after that it's looking good to me

joey-grafana (approved):

Issue #1058, opened 2026-02-09
SourceMetrics: Ability to set metric type via a prop


NWRichmond:
## Motivation

There are cases where the Metrics Drilldown app, when embedded in RCA Workbench, uses an incorrect PromQL expression because the metric lacks:
- Metadata containing the metric type
- A prefix associated with the Metrics Drilldown app's metric type heuristics

## Goal

When the `grafana-asserts-app` can determine the metric type (`gauge`, `counter`, etc.) without metric metadata or without using metric name heuristics, it would be helpful for Metrics Drilldown to receive this metric type information via a prop in the `<SourceMetrics />` exposed component. This implies some amount of reworking of metric type handling in Metrics Drilldown

bohandley:
I can see two routes to take with allowing a custom type (not defined in the Prometheus metadata and not fitting naming conventions)

1. Pass in as a URL synced variable that overrides the metadata and naming convention
2. Update the type in the function config which is saved in local storage.

Option 1 seems easily done but is surely a workaround. By coding standards it is something that would only be done for the Knowledge Graph case. There are no other areas of grafana that pass metadata, but there are other areas where naming convention is incorrect or and the metadata is undefined, which results in an incorrect `rate` being applied or not applied.

bohandley:
Here is a little more context:

### What is this?

GMD infers metric type from name suffixes (`_total`, `_sum`, `_bucket`) and falls back to `GET /api/v1/metadata?metric=<name>` when the heuristic fails. For recording rules and custom metrics this endpoint returns nothing, the negative result is not cached (cache is success-only), and the request fires again on every panel activation. GMD then treats the metric as a gauge and produces a raw-value graph where rate was expected.

KG knows the type. Let KG pass it.

### Acceptance criteria

- `<SourceMetrics />` accepts an optional `metricType: 'counter' | 'gauge' | 'histogram' | 'summary'` on each entry.
- When `metricType` is present, GMD skips the name heuristic and skips the `/api/v1/metadata` call.
- The hint propagates to the main panel and to every breakdown panel created by the Breakdown tab.
- When `metricType` is absent, GMD behaviour is unchanged.

### Implementation notes

- Thread the value through the `SourceMetrics` prop into trail construction so the `GmdVizPanel` knows the type at query-build time.
- Skip the metadata fetch path when the hint is provided.

### Out of scope

- Reworking the metadata fetch for the unhinted path. The negative-cache fix is a separate concern.

PR #1257, opened 2026-05-19
feat(sourceMetrics): per-metric `customRateInterval` override in KG source metrics component

bohandley:
### ✨ Description

**Related issue(s):** https://github.com/grafana/metrics-drilldown/issues/1130
Resolves #1130

[KG companion PR](https://github.com/grafana/asserts-app-plugin/pull/3278)

Allow KG to override `$__rate_interval` per source metric so insights derived from recording rules query at the rule's evaluation window (5m, 1h) instead of Grafana's panel-resolution-driven interval. The override flows from KG via the existing `<SourceMetrics />` exposed component and survives the embedded-to-standalone handoff via URL sync.

### 📖 Summary of the changes

- `SourceMetrics` entry type extended with three optional fields (`metricType`, `customRateInterval`, `customFunction`). Mirrors the asserts-app `AssertionSourceMetric` extension. Only `customRateInterval` is consumed in this PR; the other two are declared-only and land in their own follow-ups (#1058, #1131).
- Type extracted to `exposedComponents/SourceMetrics/types.ts` to break a `DataTrail` <-> `SourceMetrics` import cycle.
- `DataTrailState.sourceMetrics` holds the array. `SourceMetrics.tsx` passes the KG prop into `newMetricsTrail`.
- Six `GmdVizPanel` construction sites look up the active metric's entry by name and forward `customRateInterval` into `queryOptions`: `MetricGraphScene` (via a new constructor arg threaded through `MetricScene`), `MetricLabelValuesList` (single panel + per-frame Breakdown), `MetricsList`, `MetricsGroupByRow`, `ConfigurePanelForm`.
- `QueryConfig` gains `customRateInterval`. Four query builders read it with fallback to `$__rate_interval`: timeseries, stat, percentiles (histogram + non-histogram paths), heatmap.
- `DataTrail._urlSync` adds `customRateInterval`. `getUrlState` derives the value from the active metric's entry so the outbound `Open in Metrics Drilldown` link carries it. `updateFromUrl` in standalone synthesizes a single-entry `sourceMetrics` array so the same lookup path handles deep links and shared URLs. Embedded mode skips URL hydration; the React prop is authoritative.
- Per-builder tests added for the override-set case. Existing test snapshots unchanged because the absent-override path produces byte-identical PromQL.

### 🧪 How to test?

1. Pair with the asserts-app-plugin branch that extends `AssertionSourceMetric` ([linked here](https://github.com/grafana/asserts-app-plugin/pull/3278)).

Add this to KG EntityDetails.container.tsx

```
  // Temporary end-to-end smoke for metrics-drilldown issue #1130: force a 5m
  // customRateInterval on every KG source metric. useMemo keeps the array reference
  // stable so React.memo on EmbeddedMetricsDrilldown is not invalidated each render.
  const sourceMetricsWithOverride = useMemo(
    () => sourceMetrics?.map((s) => ({ ...s, customRateInterval: '5m' })),
    [sourceMetrics]
  );

...

                 {metricsQuery && (
                    <EmbeddedMetricsDrilldown
                      query={metricsQuery}
                      dataSource={metricsDataSource}
                      initialStart={drawerStart}
                      initialEnd={drawerEnd}
                      onTimeRangeChange={handleTimeRangeChange}
                      sourceMetrics={sourceMetricsWithOverride}
                      isLoadingSourceMetrics={isLoadingSourceMetrics}
                    />
                  )} 
```

2. In RCA Workbench, open a metric-bearing insight (e.g. `LatencyP99ErrorBuildup`).
3. In the embedded panel, inspect the rendered query (panel menu, Inspect, Query). Confirm `rate(metric[5m])` instead of `rate(metric[$__rate_interval])`.
4. Check the host URL contains `gmd-customRateInterval=5m`.
5. Click the `Metrics Drilldown` button. Standalone URL should contain `customRateInterval=5m` (no `gmd-` prefix). Standalone panel renders with `[5m]`.
6. Navigate to a different metric in `MetricsReducer` (multi-metric scenario) and confirm the URL drops the override when no entry matches, then restores it on back-navigation.
7. `yarn test` for the four updated query-builder suites and `DataTrail.test.tsx`.**

### DIFF
--- a/src/AppDataTrail/DataTrail.tsx
+++ b/src/AppDataTrail/DataTrail.tsx
@@ -55,6 +55,7 @@ import { getKgSceneProps, type KgEntityHint } from 'shared/knowledgeGraph/kgAnno
 import { type KgAnnotationToggle } from 'shared/knowledgeGraph/KgAnnotationToggle';
 import { logger } from 'shared/logger/logger';
 
+import { type SourceMetrics } from '../exposedComponents/SourceMetrics/types';
 import { resetYAxisSync } from '../MetricScene/Breakdown/MetricLabelsList/behaviors/syncYAxis';
 import { MetricScene } from '../MetricScene/MetricScene';
 import { type PanelDataRequestPayload } from '../shared/GmdVizPanel/components/addToDashboard/addToDashboard';
@@ -94,6 +95,10 @@ export interface DataTrailState extends SceneObjectState {
   // Knowledge Graph annotations
   kgEntityHint?: KgEntityHint;
   kgAnnotationToggle?: KgAnnotationToggle;
+
+  // Per-metric overrides supplied by KG via the SourceMetrics exposed component.
+  // Looked up by metric name at GmdVizPanel construction sites.
+  sourceMetrics?: SourceMetrics;
 
   // Add to dashboard feature
   isAddToDashboardAvailable: boolean;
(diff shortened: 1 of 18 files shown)


shelldandy (approved):
lgtm!

PR #1272, opened 2026-05-20
build(pnpm): upgrade from v10 to v11

nicwestvold:
### ✨ Description

Upgrade pnpm from v10.33.2 to v11.1.3.

pnpm 11 is a major version with a redesigned configuration system. The `pnpm` key in `package.json` is silently ignored in v11, so overrides and build dependency settings must be migrated to `pnpm-workspace.yaml`. The `pnpm/action-setup` GitHub Action also needs v6 for pnpm 11 support.

### 📖 Summary of the changes

- **`package.json`**: Removed the `pnpm` key (overrides block), bumped `packageManager` to `pnpm@11.1.3`
- **`pnpm-workspace.yaml`**: Added `overrides` block (moved from `package.json`), replaced `ignoredBuiltDependencies`/`onlyBuiltDependencies` with the new `allowBuilds` map
- **`.github/workflows/bundle-stats-create.yml`**, **`.github/workflows/is-compatible.yml`**: Bumped `pnpm/action-setup` from v5 to v6.0.8 (v5 does not support pnpm 11)

Notable new pnpm 11 defaults to be aware of:
- `minimumReleaseAge: 1440` — newly-published packages are quarantined for 24h before resolution
- `strictDepBuilds: true` — build scripts blocked unless explicitly allowed via `allowBuilds`
- `verifyDepsBeforeRun: install` — auto-runs install if deps are stale before script execution

### 🧪 How to test?

- `pnpm install`
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run test`
- `pnpm run build`
- Verify security overrides are applied: `pnpm why protobufjs`, `pnpm why dompurify`, `pnpm why postcss`, `pnpm why serialize-javascript` - all resolve to at least the override floor versions
- **Note:** The shared CI workflow (`grafana/plugin-ci-workflows@v7.3.1`) uses `pnpm/action-setup` internally and the PR to update that to v6 is [here](https://github.com/grafana/plugin-ci-workflows/pull/738). The main CI job will fail until that has been updated.

### DIFF
--- a/.github/workflows/bundle-stats-create.yml
+++ b/.github/workflows/bundle-stats-create.yml
@@ -28,7 +28,7 @@ jobs:
           echo "version=$NODE_VERSION" >> $GITHUB_OUTPUT
 
       - name: Install pnpm
-        uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5
+        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
 
       - name: Setup Node.js environment
         uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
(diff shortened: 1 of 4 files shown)


shelldandy (dismissed):
Thanks!!

shelldandy (approved):

Issue #1258, opened 2026-05-19
Epic: Handle Sea of Metrics Problem


shelldandy:
## Handle Sea of Metrics Problem

**Origin:** [#1019](https://github.com/grafana/metrics-drilldown/issues/1019) — Jan 2026 user interview
**Bet document:** [Google Doc](https://docs.google.com/document/d/1_DEMIZ5kLqWbyPL8kRk2kHpRFh8WtLuzkTsTx56snB4/edit?usp=sharing)

---

### Summary

Metrics Drilldown's existing sort options are static or usage-based and don't reflect what's happening in a system right now, leaving users to already know which metrics matter during incidents or SLO burns. This epic surfaces **runtime signals** — firing alerts, SLO burn state, and anomalies — as **filter chips**, visual indicators (icons, badges), and secondary sort options on metric cards, so impactful metrics are discoverable during triage.

#### Phase 1: Firing Alerts Filter + Visual Indicator
Highest value, lowest effort. Uses Grafana's Prometheus-compatible ruler API with server-side state filtering. All frontend, no backend.

- [ ] #1259 — Fetcher: firing alert metrics from Prometheus ruler API
- [ ] #1260 — Filter chip: "Has firing alerts"
- [ ] #1261 — Sort option: "Firing Alerts" in MetricsSorter
- [ ] #1262 — Visual indicators: alert icon + count badge on metric cards
- [ ] #1276 — Investigate Grafana Odin AB test integration for firing alert features
- [ ] #1277 — Add usage instrumentation for firing alert features
- [ ] #1278 — Validate firing alert fetcher performance at scale (1000+ rules)

#### Phase 2: SLO-Tracked Metrics
Piggybacks on SLO-generated burn-rate alerts. Conditional on `grafana-slo-app` plugin presence.

- [ ] #1263 — SLO plugin detection + burn-rate alert mapping
- [ ] #1264 — Filter chip + icon: "SLO-tracked"
- [ ] #1265 — Research: Cloud SLO API routes for richer status

#### Phase 3: Anomaly Detection (dual-track)
Cloud track: Knowledge Graph datasource for pre-computed anomaly insights. OSS track: `@bsull/augurs` WASM with a POC to validate browser performance.

- [ ] #1266 — Research: Knowledge Graph entity→metric mapping for anomaly signals
- [ ] #1267 — POC: `@bsull/augurs` anomaly detection at metric-list level
[...]

PR #1273, opened 2026-05-20
feat(sorter): add fetchFiringAlertMetrics data layer for firing alert rules

### LINKED ISSUE #1259: Fetcher: firing alert metrics from Prometheus ruler API
Part of #1258 (Handle Sea of Metrics Problem) — **Phase 1**

### What

Create `fetchFiringAlertMetrics.ts` that fetches currently firing alert rules and maps them to metric names.

### API

```
GET /api/prometheus/grafana/api/v1/rules?state=firing&limit_alerts=0
```

This is Grafana's Prometheus-compatible ruler endpoint. The `?state=firing` server-side filter reduces payload size significantly, and `limit_alerts=0` skips individual alert instance details we don't need.

### Implementation

- Follow the existing `MetricUsageFetcher` / `fetchAlertingMetrics.ts` pattern
- Parse PromQL expressions from firing rules using `extractMetricNames()`
- Return a `Map<string, number>` of `metric name → firing alert count`
- Handle errors gracefully (API unavailable, auth issues)
- Verify `?state=` query param works on minimum supported Grafana version (≥11.6.11)

### Notes

- The existing `fetchAlertingMetrics.ts` uses `/api/v1/provisioning/alert-rules` which returns rule _definitions_ only (no firing state) and is being deprecated. This new fetcher uses the non-deprecated Prometheus-compatible endpoint.
- For datasource-managed rules (Mimir/Cortex): `/api/prometheus/{ds_uid}/api/v1/rules?state=firing` has the same shape — worth considering as a stretch goal.
- No backend needed — standard Grafana HTTP API callable via `getBackendSrv()`.

### Depends on
Nothing — this is the foundational data layer for Phase 1.

### Blocks
- Filter chip: "Has firing alerts"
- Sort option: "Firing Alerts" in MetricsSorter
- Visual indicators: alert icon + count badge

shelldandy:
### ✨ Description

**Related issue(s):**
Resolves https://github.com/grafana/metrics-drilldown/issues/1259
Part of https://github.com/grafana/metrics-drilldown/issues/1258 (Handle Sea of Metrics Problem — Phase 1)

Adds `fetchFiringAlertMetrics.ts`, the foundational data layer that fetches currently firing alert rules from Grafana's Prometheus-compatible ruler endpoint and maps them to metric names. This is the first building block for firing-alert-aware features (filter chip, sort option, visual indicators).

### 📖 Summary of the changes

**New fetcher** (`src/MetricsReducer/list-controls/MetricsSorter/fetchers/fetchFiringAlertMetrics.ts`)
- Calls `GET /api/prometheus/grafana/api/v1/rules?state=firing&limit_alerts=0` via `getBackendSrv()`
- Parses the grouped Prometheus ruler response, filters to alerting rules only
- Extracts metric names from PromQL expressions using `extractMetricNames()`
- Returns `Map<string, number>` mapping each metric to the count of firing rules referencing it
- Graceful error handling — returns empty Map on API failure

**Shared constant extraction** (`fetchers/shared.ts`)
- Extracted `usageRequestOptions` (duplicated across all 3 fetchers) into a shared module
- Updated `fetchAlertingMetrics.ts` and `fetchDashboardMetrics.ts` to import from shared

**Simplifications to the new fetcher**
- Uses existing `ensureErrorObject()` from `App/errorUtils` instead of inline error coercion
- Extracted `countMetricsFromRule()` helper to reduce cognitive complexity

### 🧪 How to test?

**Unit tests** (12 tests):
```sh
npx jest fetchFiringAlertMetrics --no-coverage --forceExit
```

Covers: multi-group metric extraction, count aggregation, recording rule exclusion, empty/missing data, unparseable PromQL, network errors, string errors.

**E2E tests** (2 tests):
```sh
npx playwright test e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts --config=e2e/config/playwright.config.local.ts
```

Validates the Prometheus ruler API contract our fetcher depends on:
1. Endpoint returns provisioned alert rules with expected shape
2. `state=firing` filter returns empty groups when no alerts are firing

**Existing tests** — `fetchDashboardMetrics` tests pass unchanged (shared constant extraction is transparent).

### DIFF
--- a/e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
+++ b/e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
@@ -1,3 +1,4 @@
+import { GRAFANA_RULER_RULES_URL } from '../../../src/MetricsReducer/list-controls/MetricsSorter/fetchers/shared';
 import { expect, test } from '../../fixtures';
 
 test.describe('Firing alert metrics - Ruler API integration', () => {
@@ -7,7 +8,7 @@ test.describe('Firing alert metrics - Ruler API integration', () => {
 
   test('Ruler endpoint returns a successful response with provisioned alert rules', async ({ page }) => {
     const response = await page.request.get(
-      '/api/prometheus/grafana/api/v1/rules?limit_alerts=0'
+      `${GRAFANA_RULER_RULES_URL}?limit_alerts=0`
     );
 
     expect(response.ok()).toBe(true);
@@ -37,7 +38,7 @@ test.describe('Firing alert metrics - Ruler API integration', () => {
 
   test('Ruler endpoint with state=firing filter returns empty groups when no alerts are firing', async ({ page }) => {
     const response = await page.request.get(
-      '/api/prometheus/grafana/api/v1/rules?state=firing&limit_alerts=0'
+      `${GRAFANA_RULER_RULES_URL}?state=firing&limit_alerts=0`
     );
 
     expect(response.ok()).toBe(true);
--- a/e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
+++ b/e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
@@ -22,7 +22,11 @@ test.describe('Firing alert metrics - Ruler API integration', () => {
     const group = body.data.groups.find(
       (g: { name: string }) => g.name === 'test-evaluation-group-00'
     );
-    expect(group).toBeDefined();
+
+    if (!group) {
+      throw new Error('Expected group "test-evaluation-group-00" not found in ruler response');
+    }
+
     expect(group.rules.length).toBe(2);
 
     const ruleNames = group.rules.map((r: { name: string }) => r.name);
--- a/e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
+++ b/e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
@@ -47,6 +47,13 @@ test.describe('Firing alert metrics - Ruler API integration', () => {
 
     expect(body.status).toBe('success');
     expect(Array.isArray(body.data.groups)).toBe(true);
-    expect(body.data.groups).toHaveLength(0);
+
+    // Older Grafana versions (≤12.0.x) return group shells with empty rules arrays
+    // instead of omitting groups entirely, so assert no rules rather than no groups
+    const totalRules = body.data.groups.reduce(
+      (sum: number, g: { rules: unknown[] }) => sum + g.rules.length,
+      0
+    );
+    expect(totalRules).toBe(0);
   });
 });
--- a/e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
+++ b/e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
@@ -1,0 +1,51 @@
+import { expect, test } from '../../fixtures';
+
+test.describe('Firing alert metrics - Ruler API integration', () => {
+  test.beforeEach(async ({ metricsReducerView }) => {
+    await metricsReducerView.goto();
+  });
+
+  test('Ruler endpoint returns a successful response with provisioned alert rules', async ({ page }) => {
+    const response = await page.request.get(
+      '/api/prometheus/grafana/api/v1/rules?limit_alerts=0'
+    );
+
+    expect(response.ok()).toBe(true);
+
+    const body = await response.json();
+
+    expect(body.status).toBe('success');
+    expect(Array.isArray(body.data.groups)).toBe(true);
+    expect(body.data.groups.length).toBeGreaterThan(0);
+
+    const group = body.data.groups.find(
+      (g: { name: string }) => g.name === 'test-evaluation-group-00'
+    );
+    expect(group).toBeDefined();
+    expect(group.rules.length).toBe(2);
+
+    const ruleNames = group.rules.map((r: { name: string }) => r.name);
+    expect(ruleNames).toContain('test-rule-00');
+    expect(ruleNames).toContain('test-rule-01');
+
+    for (const rule of group.rules) {
+      expect(rule.type).toBe('alerting');
+      expect(typeof rule.query).toBe('string');
+      expect(rule.query.length).toBeGreaterThan(0);
+    }
+  });
+
+  test('Ruler endpoint with state=firing filter returns empty groups when no alerts are firing', async ({ page }) => {
+    const response = await page.request.get(
+      '/api/prometheus/grafana/api/v1/rules?state=firing&limit_alerts=0'
+    );
+
+    expect(response.ok()).toBe(true);
+
+    const body = await response.json();
+
+    expect(body.status).toBe('success');
+    expect(Array.isArray(body.data.groups)).toBe(true);
+    expect(body.data.groups).toHaveLength(0);
+  });
+});
--- a/src/MetricsReducer/list-controls/MetricsSorter/fetchers/__tests__/fetchFiringAlertMetrics.test.ts
+++ b/src/MetricsReducer/list-controls/MetricsSorter/fetchers/__tests__/fetchFiringAlertMetrics.test.ts
@@ -1,0 +1,239 @@
+import { getBackendSrv } from '@grafana/runtime';
+
+import { logger } from 'shared/logger/logger';
+
+import { fetchFiringAlertMetrics } from '../fetchFiringAlertMetrics';
+
+jest.mock('@grafana/runtime');
+jest.mock('shared/logger/logger');
+
+function setup() {
+  const get = jest.fn() as jest.Mock;
+  (getBackendSrv as jest.Mock).mockImplementation(() => ({ get }));
+  return { get };
+}
+
+function buildRulerResponse(groups: Array<{ name: string; rules: Array<Record<string, unknown>> }>) {
+  return {
+    status: 'success',
+    data: {
+      groups: groups.map((g) => ({
+        name: g.name,
+        file: 'test.yaml',
+        rules: g.rules,
+        interval: 60,
+      })),
+    },
+  };
+}
+
+function alertingRule(name: string, query: string) {
+  return { type: 'alerting', name, query, state: 'firing', health: 'ok', alerts: [], labels: {}, annotations: {} };
+}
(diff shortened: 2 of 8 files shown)


src/MetricsReducer/list-controls/MetricsSorter/fetchers/__tests__/fetchFiringAlertMetrics.test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+describe('fetchFiringAlertMetrics()', () => {
+  test('calls the correct endpoint with state=firing and limit_alerts=0', async () => {
+    const { get } = setup();
+    get.mockResolvedValueOnce(buildRulerResponse([]));
+
+    await fetchFiringAlertMetrics();
+
+    expect(get).toHaveBeenCalledWith(
+      '/api/prometheus/grafana/api/v1/rules',
```
shelldandy: this probably needs to be in a constant file and imported here

src/MetricsReducer/list-controls/MetricsSorter/fetchers/fetchFiringAlertMetrics.ts
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * Uses the `?state=firing` server-side filter to reduce payload size, and `limit_alerts=0`
+ * to skip individual alert instance details.
+ *
+ * @returns A Map of metric name → count of firing alert rules that reference the metric
+ */
+export async function fetchFiringAlertMetrics(): Promise<Map<string, number>> {
+  try {
+    const response = await getBackendSrv().get<RulerRulesResponse>(
+      '/api/prometheus/grafana/api/v1/rules',
```
shelldandy: this probably needs to be in a constant file and imported here

e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { expect, test } from '../../fixtures';
+
+test.describe('Firing alert metrics - Ruler API integration', () => {
+  test.beforeEach(async ({ metricsReducerView }) => {
+    await metricsReducerView.goto();
+  });
+
+  test('Ruler endpoint returns a successful response with provisioned alert rules', async ({ page }) => {
+    const response = await page.request.get(
+      '/api/prometheus/grafana/api/v1/rules?limit_alerts=0'
```
shelldandy: this probably needs to be in a constant file and imported here

e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    for (const rule of group.rules) {
+      expect(rule.type).toBe('alerting');
+      expect(typeof rule.query).toBe('string');
+      expect(rule.query.length).toBeGreaterThan(0);
+    }
+  });
+
+  test('Ruler endpoint with state=firing filter returns empty groups when no alerts are firing', async ({ page }) => {
+    const response = await page.request.get(
+      '/api/prometheus/grafana/api/v1/rules?state=firing&limit_alerts=0'
```
shelldandy: this probably needs to be in a constant file and imported here

e2e/tests/metrics-reducer/firing-alert-metrics.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    const body = await response.json();
+
+    expect(body.status).toBe('success');
+    expect(Array.isArray(body.data.groups)).toBe(true);
+    expect(body.data.groups.length).toBeGreaterThan(0);
+
+    const group = body.data.groups.find(
+      (g: { name: string }) => g.name === 'test-evaluation-group-00'
+    );
+    expect(group).toBeDefined();
```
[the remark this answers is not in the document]
shelldandy: Fixed in 080abf55 — replaced the `toBeDefined()` + bare access with an explicit `if (!group) throw new Error(...)` guard that fails fast with a descriptive message instead of masking the assertion with a TypeError.

src/MetricsReducer/list-controls/MetricsSorter/fetchers/fetchFiringAlertMetrics.ts
```diff
@@ -0,0 +1,108 @@
...
+  name: string;
+  file: string;
+  rules: Rule[];
+  interval: number;
+}
+
+interface Rule {
+  type: 'alerting' | 'recording';
+  name: string;
+  query: string;
```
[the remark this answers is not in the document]
shelldandy: Fixed in 080abf55 — loosened all `RulerRulesResponse`/`RuleGroup`/`Rule` fields to optional (and `query` to `unknown`) so the types reflect the defensive parsing. Type safety is restored at the call sites via a type guard filter (`rule is Rule & { name: string; query: string }`) that narrows before passing to `countMetricsFromRule`.

src/MetricsReducer/list-controls/MetricsSorter/fetchers/__tests__/fetchFiringAlertMetrics.test.ts
```diff
@@ -0,0 +1,243 @@
...
+        ])
+      );
+
+      const result = await fetchFiringAlertMetrics();
+
+      expect(result.size).toBe(1);
+      expect(result.get('up')).toBe(1);
+    });
+
+    test('skips rules with unparseable PromQL and logs a warning', async () => {
```
bohandley: Just a little nit here about this test. The promQL parser will not error on malformed promQL but on an incorrectly typed expression, like passing in `null`. The parser is built to identify errors so that they can be surfaced to the user. 

For example, `metric{}metric` promQL expr would not error, but return ['metric'] even though the query expression is incorrect.

Image

src/MetricsReducer/list-controls/MetricsSorter/fetchers/fetchFiringAlertMetrics.ts
```diff
@@ -0,0 +1,111 @@
...
+ * Uses the `?state=firing` server-side filter to reduce payload size, and `limit_alerts=0`
+ * to skip individual alert instance details.
+ *
+ * @returns A Map of metric name → count of firing alert rules that reference the metric
+ */
+export async function fetchFiringAlertMetrics(): Promise<Map<string, number>> {
+  try {
+    const response = await getBackendSrv().get<RulerRulesResponse>(
+      GRAFANA_RULER_RULES_URL,
+      { state: 'firing', limit_alerts: 0 },
```
bohandley, line 56: Do we have any idea if this needs a limit? Getting a quote from the alerting team would be a nice to have or a fuzz test with 1000 alerts or so.

bohandley (dismissed):
Solid foundation overall! Great work @shelldandy! Just a few nits left in comment. Love the stuff listed below 🥳 

  - Server-side state=firing is the right call: smaller payload, no client-side scan.
  - limit_alerts=0 correctly skips per-instance details the fetcher does not use.
  - Recording rules excluded via type === 'alerting' (line 87) per the ruler API contract.
  - Defensive Array.isArray checks (lines 77, 82) match the now-optional response interface.
  - Unit tests cover aggregation, recording-rule exclusion, and missing-field paths.
  - E2E pins the ruler contract; totalRules === 0 (vs groups.length === 0) is a thoughtful version-compat choice.
  - Request ID follows the existing fetcher naming convention.
  - shared.ts extraction is justified by real duplication across three fetchers.

shelldandy:
From the daily today:
- Lets wrap this in a feature flag
- Lets look into how to use it with Grafana Odin AB tests
- Can we instrument this? How can we measure how useful this is being/

bohandley:
> From the daily today:
> 
> * Lets wrap this in a feature flag
> * ~~Lets look into how to use it with Grafana Odin AB tests~~
> * Can we instrument this? How can we measure how useful this is being/

I think we can pause on the Odin A/B test for now! We can wrap it in a GOFF feature toggle and instrument it with rudderstack events though. We can do this in a following PR though, wrapping the feature entry point in the flag and then adding a rudderstack event for users selecting the feature. We could also use an event to see how many alerts are firing on average (this one less important but stillinteresting information).

src/MetricsReducer/list-controls/MetricsSorter/fetchers/fetchFiringAlertMetrics.ts
shelldandy: Investigated — the ruler endpoint already supports `limit` (groups), `limit_rules` (rules/group), and cursor-based pagination via `group_next_token` ([grafana/grafana#65169](https://github.com/grafana/grafana/pull/65169), [#101546](https://github.com/grafana/grafana/pull/101546)).

For our use case we intentionally want **all** firing rules to get accurate counts — adding a `limit` without pagination would give incomplete data. The `state=firing` server-side filter + `limit_alerts=0` already constrains the payload significantly (no alert instances, only firing-state rules).

Created #1278 to follow up with a scale validation: profile with 1000+ synthetic rules and get cardinality numbers from the alerting team to decide whether we need a `limit` + pagination loop as a safety valve.

src/MetricsReducer/list-controls/MetricsSorter/fetchers/__tests__/fetchFiringAlertMetrics.test.ts
shelldandy: Good catch — confirmed empirically. The lezer PromQL parser is fully tolerant of malformed string input:

```
extractMetricNames('{{{invalid promql')  → []  (no throw)
extractMetricNames('metric{}metric')     → ['metric']  (no throw)
extractMetricNames(null)                 → throws TypeError
```

It only throws on non-string input (`null`, `undefined`, `number`), which our type guard already filters out.

Fixed by:
1. Renamed the test to `'returns zero metrics for malformed PromQL that yields no identifiers'` — it now accurately describes the parser's tolerant behavior
2. Added a new test `'logs a warning and continues processing when metric extraction throws'` that mocks `extractMetricNames` to throw, proving the catch + `logger.warn` safety net works
3. Added a comment in `countMetricsFromRule` explaining the catch is a safety net since the lezer parser won't throw on strings

bohandley (approved):
LGTM! Thank you for making those changes 🚀

Issue #1261, opened 2026-05-19
Sort option: "Firing Alerts" in MetricsSorter


shelldandy:
Part of #1258 (Handle Sea of Metrics Problem) — **Phase 1**

### What

Add a new **"Firing Alerts"** entry in the sort dropdown that ranks metrics with firing alerts first, ordered by alert count descending.

### Implementation

- Extend `MetricsSorter` plugin framework with a new sort option
- Metrics with firing alerts sorted to the top, ranked by `firing alert count` descending
- Metrics without firing alerts retain their previous sort order below
- Secondary sort within the "no alerts" group: fall back to the previous/default sort

### Depends on
- #1259 (Fetcher: firing alert metrics)

PR #1281, opened 2026-05-25
feat(list-controls): add "Has firing alerts" filter chip to metrics toolbar

shelldandy:
## Summary

Browsing metrics in Metrics Drilldown showed all metrics equally — users investigating an incident had no way to surface which metrics are actively referenced by firing alert rules without already knowing the metric names. This adds a **"Has firing alerts (N)"** toggle chip to the list controls bar that narrows the metrics list to only firing-alert metrics in one click.

The chip composes with all existing filters (prefix, suffix, name search, category): the count reflects the intersection with whatever is currently active. State persists in the URL (`filter-firing-alerts=true`) for shareable filtered views. The feature is gated behind the existing `drilldown.metrics.sort_by_firing_alerts` flag.

## Design decisions

**Filter as a composable `MetricFilters` dimension, not a standalone path.** The chip plugs into the existing `EventFiltersChanged` → `MetricsReducer` → `MetricsVariableFilterEngine` pipeline by adding a `firingAlertMetrics: string[]` dimension to `MetricFilters`. When non-empty, only metrics in the firing-alert set pass through. Count computation uses the same "apply all filters except self" pattern as the sidebar sections — so the chip's displayed N updates reactively when prefix/suffix/name filters change.

**Chip always in scene graph; flag controls `visible` state.** `ListControls` builds its `SceneFlexLayout` in the `super()` call with no post-construction API for adding children. Dynamic insertion after async flag resolution would also miss `MetricsReducer.initVariablesFilteringAndSorting()`'s one-time scene-graph capture. Solution: chip is always a `SceneFlexItem` in the layout; it evaluates the flag on activation and renders `null` when the flag is off. Matches the `QuickSearch` assistant experiment pattern exactly.

**Reuse `drilldown.metrics.sort_by_firing_alerts`.** A dedicated flag for the chip would fragment Phase 1 firing-alert features. All Phase 1 work (fetcher, sort option #1261, this filter chip) shares one gate.

## Flow

```mermaid
flowchart TB
    A[FiringAlertChip activates] --> B{Flag enabled?}
    B -- No --> C[visible: false, render null]
    B -- Yes --> D[fetchFiringAlertMetrics]
    D --> E[Cache Map in state]
    E --> F{URL active?}
    F -- Yes --> G[publishEvent: EventFiltersChanged firingAlertMetrics=names]
    F -- No --> H[Idle — count 0]
    G --> I[MetricsReducer → FilterEngine.applyFilters]
    I --> J[FilteredMetricsVariable updated]
    J --> K[updateMatchingCount via VariableDependencyConfig]
```

## Testing

`FiringAlertChip.test.tsx` covers 15 cases: flag-off null render, loading spinner, toggle → correct `EventFiltersChanged` payload, `firing_alert_filter_toggled` tracking event with action and matching count, URL round-trip (`getUrlState`/`updateFromUrl`), count-0 disabled chip, and fetch-error resilience (chip stays functional, count shows 0).

`ListControls.test.ts` (new) verifies the chip is present in the scene graph at key `firing-alert-chip` without loading the full rendering stack.

`MetricsVariableFilterEngine.test.ts` extended with 7 `firingAlertMetrics` scenarios: set filter, empty passthrough, unknown names ignored, clear-and-restore, composition with prefix, composition with names, all-empty reset.

## Post-Deploy Monitoring & Validation

**Flag:** `drilldown.metrics.sort_by_firing_alerts` — chip renders `null` until the flag is on. No visible change on unflagged stacks.

**When the flag is enabled, validate:**
- Chip renders in the toolbar with `aria-label="Filter by firing alerts"`
- Toggling the chip matches the panel count to the number on the chip label
- Network tab: `reportInteraction` XHR body contains `grafana_explore_metrics_firing_alert_filter_toggled`
- URL contains `filter-firing-alerts=true` when active; pasting the URL restores the filter

**Failure signals and triggers:**
- Chip stuck on spinner → `fetchFiringAlertMetrics` silently failing; check console for `Failed to fetch firing alert rules`
- Toggle has no effect on the list → `EventFiltersChanged` not reaching `MetricsReducer`; check console for scene-graph errors
- Chip renders empty (count 0) when alerts are firing → confirm ruler endpoint returns data for the test stack (Grafana-managed rules only)


src/MetricsReducer/list-controls/FiringAlertChip/FiringAlertChip.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    try {
+      const metricsMap = await fetchFiringAlertMetrics();
+      this.setState({ firingAlertMetrics: metricsMap, loading: false });
+    } catch (err) {
+      logger.error(err instanceof Error ? err : new Error(String(err)), {
+        message: t('firing-alert-chip.fetch-error', 'FiringAlertChip: failed to load firing alert metrics'),
+      });
+      this.setState({ loading: false });
+    }
```
bohandley: Because there is a try/catch inside `fetchFiringAlertMetrics` this outer try/catch never rejects. The inside catches eveything and returns a new map. The test for this checks the failure but instead of that we could test for returning the new map.

bohandley (dismissed):
LGTM! This sorting with a highly visible button truly reduces the metrics! 

Image

Just a few nits

1. The chip looks not in line with the surrounding inputs

Image

2. Extra try/catch explained in comment

Great work and thank you for wrapping this in a feature toggle! 🥳

src/MetricsReducer/list-controls/FiringAlertChip/FiringAlertChip.tsx
shelldandy: Good catch — removed the redundant outer try/catch and updated the test to mock a resolved empty `Map` instead of a rejection, which is what actually happens when `fetchFiringAlertMetrics` fails internally. Done in 2ba0afad.

shelldandy:
Addressed both nits in 2ba0afad:

1. **Chip alignment** — Added `alignSelf: 'center'` to the chip style. The `Button size="sm"` is 24px while the toolbar's `SceneFlexItem` wrapper stretches to 32px; `alignSelf: 'center'` vertically centers it within the container.

2. **Redundant try/catch** — Removed the dead outer try/catch (replied and resolved inline).

bohandley (approved):
LGTM! 🚀

PR #1305, opened 2026-06-04
chore: remove grafana version from plugin info

bohandley:
### ✨ Description

**Related issue(s):** 
Resolves 

Remove grafana version. See here for context https://raintank-corp.slack.com/archives/C075BDBTX96/p1780488924349469

### Test

1. open plugin info and see that the specific grafana version is not listed


L2D2Grafana (approved):
👍

PR #1320, opened 2026-06-19
feat(sourceMetrics): add metricType per-entry override

### LINKED ISSUE #1058: SourceMetrics: Ability to set metric type via a prop

shelldandy:
### ✨ Description

**Related issue(s):** https://github.com/grafana/metrics-drilldown/issues/1058
Resolves https://github.com/grafana/metrics-drilldown/issues/1058

When KG sets `metricType` on a `SourceMetrics` entry, GMD skips its name-suffix heuristic and `/api/v1/metadata` fetch, using the KG hint for all type-driven decisions (rate-wrapping, panel type, aggregation, configurator presets, breakdown). When `metricType` is absent, behavior is unchanged.

**Depends on:** [asserts-app-plugin PR #3278](https://github.com/grafana/asserts-app-plugin/pull/3278) (type declaration, already merged on the asserts side).

### 📖 Summary of the changes

**Type mapping (`mapKgMetricType.ts`):** KG sends 4 Prometheus metadata values (`counter | gauge | histogram | summary`). These are mapped to GMD's 7-value internal `MetricType`:
- `counter` → `counter`, `gauge` → `gauge`, `histogram` → `classic-histogram`, `summary` → `gauge`

**Override model:** Replace — when `metricType` is present, skip both the heuristic and the metadata fetch entirely. KG only populates `metricType` on metrics where the heuristic fails (recording rules, custom metrics without standard suffixes).

**Type resolution helpers** (`getMetricType`, `getMetricTypeSync`, `getPanelTypeForMetric*`, `getConfigPresetsForMetric`): Accept optional `kgMetricType` param. When set, return the mapped type immediately.

**GmdVizPanel:** `kgMetricType` added to `QueryConfig`/`QueryOptions`. Constructor uses it for sync type resolution. `checkMetricMetadata()` skips when override is set. Histogram promotion guard still allows runtime HeatmapCells → `native-histogram` promotion when KG says `'histogram'`.

**5 construction sites** (`MetricsList`, `MetricsGroupByRow`, `MetricLabelValuesList` ×2, `ConfigurePanelForm`): Extract `entry?.metricType` alongside existing `customRateInterval`.

**4 direct call sites** (`LabelBreakdownScene`, `HistogramPercentilesDefaultBehavior`, `QueryResultsScene`/`MetricActionBar`, `getConfigPresetsForMetric`): Look up sourceMetrics entry and pass override to type helpers.

**MetricGraphScene.onActivate():** Skips the `/api/v1/metadata` fetch entirely when KG override is set.

**URL serialization:** `metricType` key added to `_urlSync`. Format: `metricType={kgType}-{metricName}`. Delimiter `-` is safe (KG types are dash-free; Prometheus names can't contain `-`). Split on first `-` to recover parts.

### 🧪 How to test?

**Unit tests (21 new):**
- `mapKgMetricType.test.ts`: all 4 mappings
- `getMetricType.test.ts`: override skips heuristic, override skips metadata fetch
- `getPanelTypeForMetric.test.ts`: override flows through
- `DataTrail.test.tsx`: URL round-trip for `metricType` key

**Manual smoke (embedded):**
1. Temp patch: add `metricType: 'counter'` to a recording-rule entry in `useAssertionSourceMetrics`
2. Open RCA Workbench for that insight
3. Confirm: no `GET /api/v1/metadata?metric=<name>` in network tab, correct `rate()` wrapping, URL contains `metricType=counter-<metricName>`

**Manual smoke (standalone deep-link):**
1. Open URL with `&metricType=counter-<metric>`
2. Confirm panel renders with counter type
3. Refresh — URL round-trips correctly

**Breakdown propagation:**
1. From a single-metric override case, open the Breakdown tab
2. Confirm every breakdown panel uses the overridden type

```
yarn test
```

### DIFF
--- a/src/AppDataTrail/DataTrail.tsx
+++ b/src/AppDataTrail/DataTrail.tsx
@@ -59,6 +59,7 @@ import { type SourceMetrics } from '../exposedComponents/SourceMetrics/types';
 import { resetYAxisSync } from '../MetricScene/Breakdown/MetricLabelsList/behaviors/syncYAxis';
 import { MetricScene } from '../MetricScene/MetricScene';
 import { type PanelDataRequestPayload } from '../shared/GmdVizPanel/components/addToDashboard/addToDashboard';
+import { type KgMetricType } from '../shared/GmdVizPanel/matchers/mapKgMetricType';
 import { MetricSelectedEvent, trailDS, VAR_DATASOURCE, VAR_FILTERS } from '../shared/shared';
 import { reportChangeInLabelFilters, reportExploreMetrics } from '../shared/tracking/interactions';
 import { buildFilterExpression } from '../shared/utils/utils.queries';
@@ -118,7 +119,7 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
   });
 
   protected _urlSync = new SceneObjectUrlSyncConfig(this, {
-    keys: ['metric', 'customRateInterval'],
+    keys: ['metric', 'customRateInterval', 'metricType'],
   });
 
   getUrlState(): SceneObjectUrlValues {
@@ -130,6 +131,7 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
     return {
       metric: this.state.metric,
       customRateInterval: entry?.customRateInterval,
+      metricType: entry?.metricType ? `${entry.metricType}-${this.state.metric}` : undefined,
     };
   }
 
@@ -146,11 +148,15 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
         ? values.customRateInterval
         : undefined;
 
-    // Standalone hydration of the KG override (issue #1130). Synthesize a single-entry
+    const parsedKgMetricType = parseKgMetricTypeFromUrl(values.metricType);
+
+    // Standalone hydration of KG overrides (issues #1130, #1058). Synthesize a single-entry
     // sourceMetrics array so the GmdVizPanel construction sites find the override via the
     // same lookup path used in embedded mode.
     const sourceMetricsOverride =
-      metric && customRateInterval ? [{ metricName: metric, labels: [], customRateInterval }] : undefined;
+      metric && (customRateInterval || parsedKgMetricType)
+        ? [{ metricName: metric, labels: [], customRateInterval, metricType: parsedKgMetricType }]
+        : undefined;
 
     this.updateStateForNewMetric(metric, sourceMetricsOverride);
   }
@@ -230,7 +236,7 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
       this.setState({
         metric,
         topScene: metric
-          ? new MetricScene({ metric, customRateInterval: entry?.customRateInterval })
+          ? new MetricScene({ metric, customRateInterval: entry?.customRateInterval, kgMetricType: entry?.metricType })
           : new MetricsReducer(),
         controls,
         ...(sourceMetricsOverride !== undefined && { sourceMetrics: sourceMetricsOverride }),
@@ -618,6 +624,20 @@ function updateAppControlsHeight() {
 
   const { height } = appControls.getBoundingClientRect();
   document.documentElement.style.setProperty('--app-controls-height', `${height}px`);
+}
+
+const VALID_KG_METRIC_TYPES = new Set(['counter', 'gauge', 'histogram', 'summary']);
+
+function parseKgMetricTypeFromUrl(raw: unknown): KgMetricType | undefined {
+  if (typeof raw !== 'string') {
+    return undefined;
+  }
+  const dashIdx = raw.indexOf('-');
+  if (dashIdx <= 0) {
+    return undefined;
+  }
+  const type = raw.slice(0, dashIdx);
+  return VALID_KG_METRIC_TYPES.has(type) ? (type as KgMetricType) : undefined;
 }
 
 function isScopesSupported(): boolean {
--- a/src/AppDataTrail/DataTrail.test.tsx
+++ b/src/AppDataTrail/DataTrail.test.tsx
@@ -158,7 +158,7 @@ describe('DataTrail - URL serialization of KG overrides', () => {
   });
 
   describe('getUrlState', () => {
-    it('emits metricType when active metric has one in sourceMetrics', () => {
+    it('emits metricType as an array of type-metric pairs', () => {
       trail.setState({
         metric: 'my_recording_rule',
         sourceMetrics: [{ metricName: 'my_recording_rule', labels: [], metricType: 'counter' }],
@@ -166,10 +166,25 @@ describe('DataTrail - URL serialization of KG overrides', () => {
 
       const urlState = trail.getUrlState();
 
-      expect(urlState.metricType).toBe('counter-my_recording_rule');
+      expect(urlState.metricType).toEqual(['counter-my_recording_rule']);
     });
 
-    it('omits metricType when entry has no metricType', () => {
+    it('emits metricType for all source metrics with overrides, not just the active one', () => {
+      trail.setState({
+        metric: 'metric_a',
+        sourceMetrics: [
+          { metricName: 'metric_a', labels: [], metricType: 'counter' },
+          { metricName: 'metric_b', labels: [], metricType: 'gauge' },
+          { metricName: 'metric_c', labels: [] },
+        ],
+      });
+
+      const urlState = trail.getUrlState();
+
+      expect(urlState.metricType).toEqual(['counter-metric_a', 'gauge-metric_b']);
+    });
+
+    it('omits metricType when no entries have metricType', () => {
       trail.setState({
         metric: 'my_metric',
         sourceMetrics: [{ metricName: 'my_metric', labels: [] }],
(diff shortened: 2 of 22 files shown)


src/AppDataTrail/DataTrail.tsx
```diff
@@ -620,6 +626,20 @@ function updateAppControlsHeight() {
   document.documentElement.style.setProperty('--app-controls-height', `${height}px`);
 }
 
+const VALID_KG_METRIC_TYPES = new Set(['counter', 'gauge', 'histogram', 'summary']);
+
+function parseKgMetricTypeFromUrl(raw: unknown): KgMetricType | undefined {
```
bohandley: Because KG can send multiple source metrics, we could have multiple type variables. Please see this for the custom function. There could be more than one custom function-metric variable in the URL so we should be able to identify each of these

https://github.com/grafana/metrics-drilldown/pull/1316/changes#r3455376770

src/AppDataTrail/DataTrail.tsx
```diff
@@ -146,11 +148,15 @@ export class DataTrail extends SceneObjectBase<DataTrailState> implements SceneO
...
+    const parsedKgMetricType = parseKgMetricTypeFromUrl(values.metricType);
+
+    // Standalone hydration of KG overrides (issues #1130, #1058). Synthesize a single-entry
     // sourceMetrics array so the GmdVizPanel construction sites find the override via the
     // same lookup path used in embedded mode.
     const sourceMetricsOverride =
-      metric && customRateInterval ? [{ metricName: metric, labels: [], customRateInterval }] : undefined;
+      metric && (customRateInterval || parsedKgMetricType)
+        ? [{ metricName: metric, labels: [], customRateInterval, metricType: parsedKgMetricType }]
+        : undefined;
```
bohandley: There is a chance that KG sends more than one metric, and so we need to be able to accept an array of mtype-metric combinations and apply them to the metrics reducer as well.

[Here is an example of where KG sends multiple source metrics](https://ops.grafana-ops.net/a/grafana-asserts-app/assertions?start=now-1h&end=now&search=Show%20insight%20PrometheusRemoteWriteDesiredShards&view=SUMMARY&ed%5Bn%5D=alloy&ed%5Btp%5D=Service&ed%5Bsc%5D%5Bns%5D=hosted-exporters&ed%5Bsc%5D%5Benv%5D=prod-us-central-0&ed%5BadditionalLabels%5D%5B__grafana_origin%5D=plugin%2Fgrafana-asserts-app&ed%5BadditionalLabels%5D%5Balertname%5D=PrometheusRemoteWriteDesiredShards&ed%5BadditionalLabels%5D%5Basserts_alert_category%5D=failure&ed%5BadditionalLabels%5D%5Basserts_entity_name%5D=alloy&ed%5BadditionalLabels%5D%5Basserts_entity_type%5D=Service&ed%5BadditionalLabels%5D%5Basserts_env%5D=prod-us-central-0&ed%5BadditionalLabels%5D%5Basserts_severity%5D=warning&ed%5BadditionalLabels%5D%5Bcluster%5D=prod-us-central-0&ed%5BadditionalLabels%5D%5Bcomponent_id%5D=%5B%20%22prometheus.remote_write.stack_289729_cloudwatch_notags%22,%20%22prometheus.remote_write.stack_291579_cloudwatch_notags%22,%20%22prometheus.remote_write.stack_326901_cloudwatch_notags%22,%20%22prometheus.remote_write.stack_436716_cloudwatch_tags%22%20%5D&ed%5BadditionalLabels%5D%5Bcomponent_path%5D=%2Fcloudwatch_pipelines.pipeline_cloudwatch&ed%5BadditionalLabels%5D%5Bcontainer%5D=grafana-agent&ed%5BadditionalLabels%5D%5Binstance%5D=%5B%20%22alloy-0:grafana-agent:http-metrics%22,%20%22alloy-1:grafana-agent:http-metrics%22,%20%22alloy-2:grafana-agent:http-metrics%22,%20%22alloy-3:grafana-agent:http-metrics%22,%20%22alloy-4:grafana-agent:http-metrics%22,%20%22alloy-5:grafana-agent:http-metrics%22%20%5D&ed%5BadditionalLabels%5D%5Bjob%5D=hosted-exporters%2Falloy&ed%5BadditionalLabels%5D%5Bns%5D=hosted-exporters&ed%5BadditionalLabels%5D%5Bpod%5D=%5B%20%22alloy-0%22,%20%22alloy-1%22,%20%22alloy-2%22,%20%22alloy-3%22,%20%22alloy-4%22,%20%22alloy-5%22%20%5D&ed%5BadditionalLabels%5D%5Bprovider%5D=google&ed%5BadditionalLabels%5D%5Bremote_name%5D=%5B%20%2206c96c%22,%20%2225ead4%22,%20%2235d4d4%22,%20%226c1b26%22,%20%22772544%22,%20%22778f81%22,%20%227d097c%22,%20%22c31a91%22,%20%22cb502b%22,%20%22e4e479%22%20%5D&ed%5BadditionalLabels%5D%5Bservice_type%5D=agent&ed%5BadditionalLabels%5D%5Bteam%5D=asserts&ed%5BadditionalLabels%5D%5Btenant%5D=27821&ed%5BadditionalLabels%5D%5Burl%5D=http:%2F%2Fcortex-gw-internal.cortex-prod-10.svc.cluster.local.%2Fapi%2Fprom%2Fpush&ed%5BalertName%5D=PrometheusRemoteWriteDesiredShards&ed%5Btab%5D=metricsView&ed%5Bstart%5D=now-1h&ed%5Bend%5D=now&gmd-filters-prefix=prometheus&gmd-var-metrics-reducer-sort-by=alphabetical&from=now-1h&to=now&timezone=utc&gmd-var-ds=grafanacloud-prom&gmd-var-metrics_filters=asserts_env%7C%3D%7Cprod-us-central-0&gmd-var-metrics_filters=__name__%7C%3D~%7C%28prometheus_remote_storage_shards_desired__gfp__prometheus_remote_storage_shards_max%29&gmd-var-metrics_filters=namespace%7C%3D%7Chosted-exporters&gmd-var-metrics_filters=job%7C%3D%7Chosted-exporters%2Falloy&gmd-var-filters=asserts_env%7C%3D%7Cprod-us-central-0&gmd-var-filters=__name__%7C%3D~%7C%28prometheus_remote_storage_shards_desired__gfp__prometheus_remote_storage_shards_max%29&gmd-var-filters=namespace%7C%3D%7Chosted-exporters&gmd-var-filters=job%7C%3D%7Chosted-exporters%2Falloy&gmd-var-labelsWingman=%28none%29&gmd-layout=grid&gmd-filters-rule=&gmd-filters-suffix=&gmd-search_txt=&gmd-filters-recent=&gmd-var-other_metric_filters=filters-prefix,prefix%7C%3D%7Cprometheus)

bohandley (changes requested):
Looks like a good start! We need to change the implementation a little to handle multiple metricType variables and apply the types for metric reducer and for the metric scene.

Some KG insights always have multiple source metrics like `PrometheusRemoteWriteDesiredShards` insight. This one will always show multiple metrics in the reducer. 
Image

I think we'll need to merge https://github.com/grafana/metrics-drilldown/pull/1316 before this one so you can leverage the multi metric variable functionality. Let me know if you have any questions!

src/AppDataTrail/DataTrail.tsx
shelldandy: Good call — I've refactored this to use the same multi-value URL parsing pattern from #1316. The new `parseMetricTypeValues()` in `sourceMetricsUrlSync.ts` normalises both single-string and array inputs into a `Map<metricName, KgMetricType>`, with validation against the four valid types. The old `parseKgMetricTypeFromUrl` is removed.

src/AppDataTrail/DataTrail.tsx
shelldandy: Addressed — rebased onto main (picking up #1316), and extended `buildSourceMetricsOverride` to accept a `metricTypeByMetric` map alongside `customFunctionByMetric`. The URL now serialises `metricType` as a multi-value array (`metricType=counter-metric_a&metricType=gauge-metric_b`), so all source metrics with type overrides survive the embedded→standalone handoff. All downstream consumption sites already look up per-metric from the `sourceMetrics` array, so no changes needed there.

bohandley (approved):
LGTM! Thank you for this work @shelldandy! This delivers the functionality for allowing KG to pass custom types for multiple metrics to GMD and preserve those types in the URL.

https://github.com/user-attachments/assets/97c0e102-37bf-4d59-9a55-f084a343fa87

Two non blocking nits that we can do a fast follow for:

1. If there are multiple `metricType` vars in the URL and this URL is pasted into a new window, DataTrail only reads the first metricType and drops the second. This is not a breaking change for the KG use case, but more of a nice to have if people begin sharing the URL with the custom types. I think this also aplies to the custom function PR as well.
2. We can clean up the repeated Prometheus types which are repeated in `VALID_KG_METRIC_TYPES`, `KgMetricType` and `mapKgMetricType`. I know the last one handles the special classic-histogram naming, but I think there is a way to clean this up without repeating the types 3 times.

Great work, thank you for delivering this! 🥳

PR #1323, opened 2026-06-23
fix(config-presets): use resolved metric type instead of re-deriving …

### LINKED ISSUE #1228: Configure Prometheus function dialog shows wrong presets for native histograms
## Summary
The "Configure the Prometheus function" drawer determines which presets to show (Heatmap, Percentiles, Sum, Avg, etc.) by re-deriving the metric type from Prometheus `/metadata`. Prometheus metadata is unreliable for native histograms — it frequently returns no entry or a non-`histogram` type — so the drawer falls back to the gauge preset list and the **Heatmap** option never appears for those metrics.
This is the same root cause that PR #1207 fixed for the main viz panel by switching to data-frame–based detection. That fix did not extend to the configure drawer, so the drawer is still metadata-only.

## Repro
1. Open Metrics Drilldown for a native-histogram metric whose Prometheus `/metadata` does not return `type: histogram`. Example we hit this on: `adaptive_logs_get_exemptions_success_latency_seconds`.
2. Wait for the main panel to render. With #1207, the main panel correctly identifies it as a native histogram via the data frame and renders as a heatmap.
3. Click the "Configure Prometheus function" cog in the panel header.
4. **Expected:** the drawer shows the histogram presets (`Heatmap (default)`, `Percentiles`).
5. **Actual:** the drawer shows the gauge presets (`Average`, `Sum`, `Standard deviation`, `Percentiles`, `Minimum and maximum`). The drawer subtitle reads `<metric> (gauge)`.
For comparison, classic histograms (e.g. `adaptive_logs_api_request_duration_seconds_bucket`) work correctly because they are detected synchronously from the `_bucket` suffix.

## Root cause
`ConfigurePanelForm.buildBody` passes only `metric.name` and re-derives the type:

```ts
// src/shared/GmdVizPanel/components/ConfigurePanelForm/ConfigurePanelForm.tsx
const presets = await getConfigPresetsForMetric(metric.name, getTrailFor(this));
```

`getConfigPresetsForMetric` then asks `getMetricType`, which is metadata-only:

```ts
// src/shared/GmdVizPanel/config/presets/getConfigPresetsForMetric.ts
export async function getConfigPresetsForMetric(metric: string, dataTrail: DataTrail): Promise<PanelConfigPreset[]> {
  const metricType = await getMetricType(metric, dataTrail);
  switch (metricType) {
    case 'classic-histogram':
    case 'native-histogram':
      return Object.values(DEFAULT_HISTOGRAMS_PRESETS);
    ...
    default:
      return Object.values(DEFAULT_TIMESERIES_PRESETS);
  }
}
```

```ts
// src/shared/GmdVizPanel/matchers/getMetricType.ts
if (metricType === 'gauge') {
  const metadata = await dataTrail.getMetadataForMetric(metric);
  if (metadata?.type === 'histogram') {
    return 'native-histogram';
  }
  ...
}
```

So when metadata doesn't say `histogram`, the drawer falls into the `default` (timeseries/gauge) branch regardless of what the data frames actually look like.
By the time the user clicks the cog, the surrounding `MetricGraphScene` has already updated `ConfigurePanelAction`'s `metric.type` to `'native-histogram'` (via the data-frame subscription in `GmdVizPanel`), but the dialog ignores that resolved type and re-asks metadata.

## Why PR #1207 alone is not enough

#1207 moved the main panel off of metadata for native-histogram detection, which is correct. But the configure drawer still relies on metadata via `getConfigPresetsForMetric` → `getMetricType`. So the drawer remains broken on exactly the metrics #1207 was motivated by.

### Option 1 — Trust the resolved `Metric.type` (smaller change)
Change `getConfigPresetsForMetric` to take the already-resolved metric type instead of re-deriving it, and have `ConfigurePanelForm` pass it through:

```ts
// getConfigPresetsForMetric.ts — sketch
export function getConfigPresetsForMetric(metricType: MetricType): PanelConfigPreset[] {
  switch (metricType) { ... }
}
// ConfigurePanelForm.tsx — sketch
const presets = getConfigPresetsForMetric(metric.type);
```

By the time the cog is clickable and clicked, `metric.type` is the value that `MetricGraphScene` already corrected from data frames. This is consistent with the direction of #1207.

**Caveat:** if the user clicks the cog before the main panel's first query completes, `metric.type` will still be the synchronous heuristic (`'gauge'`) and the drawer will show gauge presets. We may want to disable or delay the cog until the main panel has resolved its type, or fall back to Option 2 for robustness.

### Option 2 — Drawer-side data-frame probe (more robust)
Have `ConfigurePanelForm` subscribe to the same data-frame signal `GmdVizPanel` uses, and rebuild its preset list when `HeatmapCells` is seen. This removes any dependency on click timing or `MetricGraphScene` having already corrected the type, but is a larger change.

## Acceptance criteria
- For a native histogram with broken/missing Prometheus metadata, the configure drawer shows the histogram presets (Heatmap default + Percentiles).
- The drawer subtitle reads `<metric> (native-histogram)`.
- Behavior unchanged for classic histograms, gauges, counters, etc.
- Tests cover the case where `getMetricType` would return `'gauge'` but the metric is actually a native histogram (i.e., metadata disagrees with reality).

## Related
- PR #1207 — fix(native-histogram): rely solely on data frames for native histogram detection (covers the main viz panel only)

AvinashDevX:
…from metadata

Fixes #1228. The 'Configure Prometheus function' drawer's preset list was re-deriving the metric type via getMetricType() (metadata-only), which for native histograms often returns no entry or non-'histogram' type and so falls back to the gauge presets. The drawer is now passed the already- resolved MetricType from ConfigurePanelForm.state.metric.type, which has already benefited from the data-frame-based detection introduced in #1207.

- getConfigPresetsForMetric: now takes MetricType directly (sync)
- ConfigurePanelForm.buildBody: passes metric.type
- Adds regression test covering native-histogram + the other branches

### ✨ Description

**Related issue(s):** 
Resolves


shelldandy (changes requested):
@AvinashDevX could you sign the CLA please? also seems like there are conflicts

AvinashDevX:
Thanks for the review. Conflicts are resolved and the CLA is signed.

shelldandy (approved):

shelldandy:
@AvinashDevX could you amend your commits to include a gpg signature.
We need verified commits as part of our security protocols.

I think after that it's looking good to me

bohandley:
Sorry @AvinashDevX and @shelldandy, we have to close this because the post data response analysis needed to be fixed. The data response check(that this PR relies on) for heatmap cells failed on native histograms by returning no data (without a rate function) so we had to come up with another solution. The solution that was just merged is a pre-query on gauges without metadata, where we run a probe test with `sum(rate(<no metadata gauge metric>))`. This will return heatmap cells if it is a native histogram and then we are certain.

I'm working on a refactor of that right now that should land in the next day. Thank you for understanding!

Issue #1277, opened 2026-05-21
Add usage instrumentation for firing alert features


shelldandy:
## Context

Part of #1258 (Handle Sea of Metrics Problem — Phase 1)

Raised during daily standup review of #1273 (fetchFiringAlertMetrics data layer): "Can we instrument this? How can we measure how useful this is being?"

## Goal

Add analytics/instrumentation so we can measure adoption and usefulness of the firing alert features (sort, filter chip, visual indicators).

## Tasks

- [ ] Track when users select the "Firing Alerts" sort option
- [ ] Track when users toggle the "Has firing alerts" filter chip
- [ ] Track fetch success/failure rates and response times for the ruler API call
- [ ] Track result set sizes (how many metrics have firing alerts in a typical session)
- [ ] Emit events compatible with the existing Grafana analytics pipeline
- [ ] Define a dashboard or data source for reviewing these signals

## Notes

This instrumentation should be compatible with the Odin AB test work (#1276) so we can compare metrics between control and variant groups.

Issue #1262, opened 2026-05-19
Visual indicators: alert icon + count badge on metric cards


shelldandy:
Part of #1258 (Handle Sea of Metrics Problem) — **Phase 1**

### What

Show a **tooltip icon** (bell or flame) on metric cards that have firing alerts, plus an **alert count badge** visible on hover.

### Design

- Icon is subtle — avoids the "everything is highlighted = nothing is highlighted" problem seen with KG annotations
- Alert count badge shown on hover or alongside the icon
- Visible regardless of active filters or sort order (orthogonal to filter/sort)
- Tooltip on the icon could show alert rule names or a summary

### Depends on
- #1259 (Fetcher: firing alert metrics)

Issue #1260, opened 2026-05-19
Filter chip: "Has firing alerts"


shelldandy:
Part of #1258 (Handle Sea of Metrics Problem) — **Phase 1**

### What

Add a **"Has firing alerts"** filter chip alongside existing label filters. This is the **primary interaction** surface for runtime signals — users toggle it to see only metrics with currently firing alerts.

### Design

- Quick-filter toggle chip, visually consistent with existing label filter chips
- Binary filter: a metric either has firing alerts or it doesn't
- When active, only metrics present in the firing alerts map are shown
- Chip should display count of matching metrics (e.g., "Has firing alerts (7)")

### Why filter over sort

Filter is the right UX for binary signals. Sort only adds value when there's a meaningful hierarchy (e.g., alert count ranking), which is handled separately in the sort option issue.

### Depends on
- #1259 (Fetcher: firing alert metrics)

PR #1342, opened 2026-07-08
feat(sea-of-metrics): Phase 1 — firing alerts sort, visual badges, and instrumentation

### LINKED ISSUE #1261: Sort option: "Firing Alerts" in MetricsSorter

shelldandy:
### ✨ Description

**Related issue(s):** https://github.com/grafana/metrics-drilldown/issues/1258
Resolves #1261, #1262, #1277

Phase 1 continuation of the Sea of Metrics epic. Builds on top of the existing fetcher (#1259) and filter chip (#1260) to add the remaining firing alert features: sort option, visual indicators on cards, and usage instrumentation.

All features are gated behind the `drilldown.metrics.sort_by_firing_alerts` feature flag.

### 📖 Summary of the changes

**Cached firing-alert provider** — Added `getFiringAlertCounts()`, `getFiringAlertCountForMetric()`, and `getFiringAlertCountsAsRecord()` to `MetricsSorter` with request deduplication and caching, mirroring the existing `MetricUsageFetcher` pattern. All consumers (`FiringAlertBadge`, `FiringAlertChip`, `MetricsVariableSortEngine`) route through `MetricsSorter` to ensure only one ruler API call and one tracking event fire regardless of how many consumers need the data.

**Sort option (#1261)** — Added `'firing-alerts'` to `SortingOption` type. The option is included in the initial `CustomVariable` query and removed at activation time when the flag is off (reverse gating — `CustomVariable` doesn't support dynamic option injection after activation, so `refreshOptions()` is called to re-render the dropdown after removal). Generic unsupported-value migration runs synchronously at activation; only the firing-alerts removal is gated on the flag promise. `MetricsVariableSortEngine` gets a new `sortByFiringAlerts()` method that sorts metrics descending by firing alert count using the existing `sortMetricsByCount()` utility.

**Visual indicators (#1262)** — New `FiringAlertBadge` SceneObjectBase component renders a bell icon with count as a `headerAction` on metric cards. Self-gates on the feature flag and uses `MetricsSorter.getFiringAlertCountForMetric()` for O(1) cached lookups. Added to both `MetricsList` and `MetricsGroupByRow` card rendering.

**Instrumentation (#1277)** — Added `performance.now()` timing to `fetchFiringAlertMetrics()` and a new `firing_alert_metrics_fetched` tracking event reporting `status`, `duration_ms`, `metric_count`, and `rule_count`. The existing `sorting_changed` event automatically captures `sortBy: 'firing-alerts'` via `MetricsReducer`'s `EventSortByChanged` subscription.

**Grafana feature toggle override for GoFF flags** — Added a `featureToggle` field to GoFF flag definitions that maps to a Grafana server-side feature toggle. When the toggle is set via `GF_FEATURE_TOGGLES_ENABLE`, `evaluateFeatureFlag()` returns its value directly without consulting the OpenFeature provider. This eliminates the need to patch source files for local dev/QA — the firing alerts flag maps to `metricsExploreFireAlerts`.

**Dev tooling** — Added `pnpm server:firing-alerts` script that passes `GF_FEATURE_TOGGLES_ENABLE=metricsExploreFireAlerts` to docker compose. Also added 3 always-firing provisioned alert rules (`provisioning/alerting/alert-rules-wingman.json`) so firing alerts data is available immediately without manual rule creation.

### 🧪 How to test?

**Local setup:**
```bash
# Terminal 1
pnpm dev

# Terminal 2 — starts Grafana with the flag enabled + always-firing alert rules provisioned
pnpm server:firing-alerts
```

Then visit http://localhost:3001/a/grafana-metricsdrilldown-app/ — the provisioned alert rules fire within ~1 minute.

**Sort option:**
- Open the sort dropdown → "Firing Alerts" should appear as the last option
- Select it → metrics with firing alerts sort to the top, ordered by alert count descending
- Copy the URL with `firing-alerts` sort active, open in new tab → sort should persist
- Non-alerting metrics appear below, sorted alphabetically

**Visual badges:**
- Metric cards with firing alerts show a bell icon + count badge in the card header
- Hover the badge → tooltip shows "N firing alert rule(s) reference this metric"
- Badges appear in both grid and group-by views
- Metrics without firing alerts do NOT show a badge

**Filter chip + sort combo:**
- Activate "Has firing alerts" chip → filters to only alerting metrics
- With "Firing Alerts" sort active simultaneously → both compose correctly

**Instrumentation:**
- DevTools → Console: `grafana_explore_metrics_firing_alert_metrics_fetched` events with `status`, `duration_ms`, `metric_count`, `rule_count`
- Selecting "Firing Alerts" sort emits `grafana_explore_metrics_sorting_changed` with `sortBy: 'firing-alerts'`

**Flag off (regression):**
- Run `pnpm server` (normal, no flag override) → sort option should not appear, no badges, chip hidden
- URL with `var-metrics-reducer-sort-by=firing-alerts` and flag off → migrates to `default`

### Screenshots

image

image

image


src/MetricsReducer/list-controls/MetricsSorter/MetricsSorter.tsx
```diff
@@ -171,6 +190,29 @@ export class MetricsSorter extends SceneObjectBase<MetricsSorterState> {
...
+    }
+    this.firingAlertCache.promise = fetchFiringAlertMetrics().then((data) => {
+      this.firingAlertCache.data = data;
+      this.firingAlertCache.promise = null;
+      return data;
+    });
+    return this.firingAlertCache.promise;
+  }
+
+  public getFiringAlertCountForMetric(metric: string): Promise<number> {
```
bohandley, line 212: This gets called in the firing alerts badge code on main already, so the claim of the PR description that there is one call is inaccurate. We now call this twice so I believe the rudderstack event will get fired twice as well resulting in inaccurate usage data.

Can we route the chip through MetricsSorter.getFiringAlertCounts() the same way FiringAlertBadge does?

src/MetricsReducer/list-controls/MetricsSorter/MetricsSorter.tsx
```diff
@@ -142,10 +150,21 @@ export class MetricsSorter extends SceneObjectBase<MetricsSorterState> {
...
-    }
+    evaluateFeatureFlag('drilldown.metrics.sort_by_firing_alerts').then((enabled) => {
+      if (!enabled) {
+        this.supportedSortByOptions.delete('firing-alerts');
+        const query = sortByVar.state.query
+          .split(',')
+          .filter((part) => !part.includes('firing-alerts'))
+          .join(',');
+        sortByVar.setState({ query });
+        sortByVar.validateAndUpdate();
```
bohandley: `validateAndUpdate` is an optional function on the variable and since we don't define it, this is a no op. I would suggest using refreshOptions() instead.

src/MetricsReducer/list-controls/MetricsSorter/MetricsSorter.tsx
```diff
@@ -142,10 +150,21 @@ export class MetricsSorter extends SceneObjectBase<MetricsSorterState> {
...
+      }
+
+      if (!this.supportedSortByOptions.has(sortByVar.getValue() as SortingOption)) {
```
bohandley: Before this PR, if (!supportedSortByOptions.has(...)) changeValueTo('default') ran synchronously in the activation handler. Now it runs inside evaluateFeatureFlag(...).then(...).

Can we keep the generic unsupported-value migration synchronous, and only gate the firing-alerts removal on the flag promise?

bohandley (changes requested):
Looking great! Just a few requested changes mentioned in the comments. 🥳 

Two things to consider moving forward after this PR

1. Do we want to consider a time limit to the cache? Alerts may be not be firing and then start firing at any point in time. I would suggest a follow up for cache invalidation, with a little research on choosing a time.
2. I love the badge as an eye catching UI component for a new feature. Have we thought how long we keep this and if this badge functionality will move to the sidebar, as a metrics filter? I would suggest keeping the badge for at least a month, then moving to the sidebar. I like the idea of adding new filter features explicitly and clearly outside of the sidebar, but I am hesitant to squeeze in all the future filters in this location. What will be the next filter feature to reduce the sea of metrics?

src/MetricsReducer/list-controls/MetricsSorter/MetricsSorter.tsx
shelldandy: Good catch — fixed in 7e537e1d. `FiringAlertChip` now uses `sceneGraph.findByKeyAndType(this, 'metrics-sorter', MetricsSorter)` and calls `metricsSorter.getFiringAlertCounts()`, same pattern as `FiringAlertBadge`. This deduplicates both the network request and the Rudderstack event. Updated the PR description as well.

src/MetricsReducer/list-controls/MetricsSorter/MetricsSorter.tsx
shelldandy: Fixed in 7e537e1d — replaced `validateAndUpdate()` with `refreshOptions()`.

src/MetricsReducer/list-controls/MetricsSorter/MetricsSorter.tsx
shelldandy: Good point — fixed in 7e537e1d. The generic unsupported-value check now runs synchronously before the `evaluateFeatureFlag` call. The flag callback only handles the `firing-alerts`-specific removal and re-checks if the current value is specifically `firing-alerts` afterward.

shelldandy:
Thanks for the review @bohandley! All three inline comments are addressed in 7e537e1d.

On the two forward-looking points:

1. **Cache TTL** — Agreed, a time-based invalidation makes sense. The current cache lives for the `MetricsSorter` lifetime with no expiry, so a user who stays on the page for a while won't see newly firing alerts. I'll create a follow-up to research a reasonable TTL (likely in the 2–5 min range to balance freshness vs. API load).

2. **Badge lifecycle & filter location** — The badge is intended as a discovery mechanism for the first phase. I agree with keeping it for at least a month and then evaluating whether the firing alerts filter moves into the sidebar as a permanent filter option. For next filter candidates, I'm thinking metric type (counter/gauge/histogram) or label-based grouping, but happy to discuss further in a separate issue.

bohandley (approved):
LGTM! Let's create an issue to track caching invalidation and adjusting the badge after a time. Thank you for this work @shelldandy 🚀

PR #1349, opened 2026-07-20
fix(faro): deduplicate InlineBanner error reporting to prevent SLO breach

### LINKED ISSUE #1348: Investigate breaching and misconfigured SLOs
## Context

Weekly health check (2026-07-20) flagged two SLO issues requiring follow-up.

## 1. Exception-free session rate SLO is critically breaching

- **SLO UUID:** `vz0shxu08bxqlp2us5czb`
- **Current SLI:** -49.5% (objective: 80%)
- **Burn rate:** 4.96x
- **Error budget:** -647% (fully exhausted)
- **Trend:** Steadily declining over 28 days — from ~34% on Jun 22 to -49% today

The SLO formula is `(sessions - exceptions) / sessions` using Faro collector metrics (`faro_collector_sessions_total` and `faro_collector_app_exceptions_count` for appId 49). A negative SLI means the exception count far exceeds the session count — on average, each session is generating multiple exceptions.

### Suggested investigation

- Check [Kowalski errors for app 49](https://ops.grafana-ops.net/a/grafana-kowalski-app/apps/49/errors?from=now-7d&to=now&timezone=utc&frameIndex=0) to identify the top error groups
- Determine if a specific noisy/spammy exception is flooding the count, or if there's a Faro instrumentation issue inflating `faro_collector_app_exceptions_count`
- Consider whether the SLO formula needs refinement (e.g., deduplicating exceptions per session, or excluding known non-actionable exceptions)

### SLO link

https://ops.grafana-ops.net/a/grafana-slo-app/manage-slos?labels=service_name%3Dgrafana-metricsdrilldown-app

## 2. Logs Drilldown Plugin Loading SLO shows NODATA

- **SLO UUID:** `10aewxg1v9790hg6hzq6r`
- **Objective:** 99.5% over 28d
- **Status:** NODATA — recording rules are not producing metrics

### Suggested investigation

- Check the SLO definition's destination datasource configuration
- Verify that the recording rules are being evaluated and writing to the expected Prometheus

shelldandy:
## Summary

- Move `logger.error()` from `InlineBanner`'s render body into a `useEffect` with a ref guard, so each unique error is reported to Faro exactly once instead of on every React re-render
- Add test coverage for `InlineBanner` error deduplication behavior

## Root cause

The **Exception-free session rate** SLO is critically breaching at **-49.5%** (objective: 80%). Over the last 28 days: **771k exceptions vs 528k sessions** — a 1.46:1 ratio, with spikes to 3.53x.

The SLO formula is `(sessions - exceptions) / sessions`. The primary amplifier is `InlineBanner`, which calls `logger.error()` → `faro.api.pushError()` directly in the render function body. Every React re-render of a component displaying an error pushes a new Faro exception. A single query failure can generate dozens of exceptions per session as React re-renders.

`InlineBanner` with the `error` prop is used in 7+ locations across the app (MetricsList, MetricLabelsList, MetricLabelValuesList, MetricsGroupByRow, MetricsGroupByList, QueryResultsScene, ErrorView).

## Test plan

- [x] `logger.error` is called exactly once on initial render with an error
- [x] Re-rendering with the same error does NOT call `logger.error` again
- [x] Changing the error prop DOES call `logger.error` with the new error
- [x] Rendering without an error prop does NOT call `logger.error`
- [x] Error message formatting (including HTTP status) renders correctly
- [x] Typecheck passes
- [x] Lint passes

After deploy, monitor the SLO timeline — the exceptions-per-session ratio should drop below 1.0 within 24-48h.

Closes #1348

### DIFF
--- a/src/App/InlineBanner.test.tsx
+++ b/src/App/InlineBanner.test.tsx
@@ -1,0 +1,84 @@
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+
+import { InlineBanner } from './InlineBanner';
+import { logger } from '../shared/logger/logger';
+
+jest.mock('@grafana/i18n', () => ({
+  ...jest.requireActual('@grafana/i18n'),
+  t: (_key: string, defaultValue: string) => defaultValue,
+}));
+
+const mockLogger = logger as jest.Mocked<typeof logger>;
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+describe('InlineBanner', () => {
+  it('renders without error prop', () => {
+    render(<InlineBanner severity="info" title="Info" message="All good" />);
+
+    expect(screen.getByText('Info')).toBeInTheDocument();
+    expect(screen.getByText('All good')).toBeInTheDocument();
+    expect(mockLogger.error).not.toHaveBeenCalled();
+  });
+
+  it('logs error exactly once on initial render', () => {
+    const error = new Error('query failed');
+
+    render(<InlineBanner severity="error" title="Error" error={error} />);
+
+    expect(mockLogger.error).toHaveBeenCalledTimes(1);
+    expect(mockLogger.error).toHaveBeenCalledWith(
+      expect.objectContaining({ message: 'query failed' }),
+      expect.objectContaining({ bannerTitle: 'Error' })
+    );
+  });
+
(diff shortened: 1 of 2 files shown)


bohandley (approved):
Good catch! 🚀

PR #1357, opened 2026-07-24
feat(sea-of-metrics): make firing-alerts flag an Odin A/B test experiment

shelldandy:
### ✨ Description

**Related issue(s):** Follow-up to #1342 (Phase 1 — firing alerts sort, badges, and instrumentation)

Prepares the "sort by firing alerts" feature (shipped in #1342 behind `drilldown.metrics.sort_by_firing_alerts`) to be run as a gradual-release **A/B test with Odin**.

The KPI instrumentation from #1342 was already wired correctly — the gap for Odin was that the flag was a plain boolean with no experiment cohort, so the firing-alerts KPI events couldn't be correlated to a treatment/control group. This PR converts the flag to the same treatment/control/excluded cohort pattern already used by our `default_open_sidebar` and assistant experiments, and enriches the firing-alerts KPI events with the cohort value (Odin "Single Query" correlation).

> **Note:** This is the in-repo half. Two out-of-repo steps follow separately: (1) the `grafana/deployment_tools` GoFF libsonnet flag definition (analogous to `default-open-sidebar.libsonnet`) — a hard dependency for cohorts to be served in prod, and (2) the KPI dashboard panels + Odin experiment setup.

### 📖 Summary of the changes

- **`openFeature.ts`** — convert `drilldown.metrics.sort_by_firing_alerts` from `boolean` to a `string` cohort flag (`treatment`/`control`/`excluded`, default `excluded`) with `trackingKey: experiment_sort_by_firing_alerts`. Add an `isFiringAlertsSortingEnabled()` helper (only `treatment` enables the feature) so the treatment→enabled rule lives in one place.
- **3 consumers** — `MetricsSorter`, `FiringAlertChip`, and `FiringAlertBadge` now route through the helper instead of checking flag truthiness directly (`control`/`excluded` both hide the feature).
- **`tracking.ts`** — register the `experiment_sort_by_firing_alerts` cohort slot in `TRACKED_FLAG_VALUES`.
- **`interactions.ts`** — `getExperimentPayloads` now enriches `firing_alert_filter_toggled`, `firing_alert_metrics_fetched`, and `sorting_changed` (only when `sortBy === 'firing-alerts'`) with the cohort value.
- **Local dev preserved** — the `metricsExploreFireAlerts` feature-toggle override (used by `pnpm server:firing-alerts`) still works: for string cohort flags the boolean toggle maps to `treatment`/`control`. Extracted this into a small `resolveFeatureToggleOverride` helper.
- **Tests** — updated the 3 consumer suites to mock the new helper, and added `openFeature.test.ts` cases for the string-flag default (`excluded`) and the dev-toggle→cohort mapping.

### 🧪 How to test?

- `pnpm typecheck` and `pnpm test` pass (full suite: 734 tests). Targeted suites: `openFeature.test`, `MetricsSorterFiringAlerts`, `FiringAlertChip`, `FiringAlertBadge`.
- Local smoke: `pnpm server:firing-alerts` (maps the dev toggle to `treatment`) → confirm the "Firing Alerts" sort option, the "Has firing alerts" filter chip, and metric alert badges all render.
- In browser devtools → Network, confirm the `reportInteraction` payloads for `sorting_changed` (with `sortBy=firing-alerts`) and `firing_alert_filter_toggled` include the `experiment_sort_by_firing_alerts` field.


src/shared/featureFlags/openFeature.ts
```diff
@@ -188,3 +217,15 @@ export async function evaluateFeatureFlag<T extends keyof typeof goffFeatureFlag
...
+
+/**
+ * Whether the "sort by firing alerts" experiment feature (firing-alerts sort option, "Has firing alerts" filter
+ * chip, and metric alert badges) should be shown to the current user.
+ *
+ * The `drilldown.metrics.sort_by_firing_alerts` flag is an A/B test with `treatment`/`control`/`excluded`
+ * cohorts; only the `treatment` cohort sees the feature. Evaluating the flag also records the cohort for
+ * analytics enrichment (see {@link TrackingHook}).
+ */
+export async function isFiringAlertsSortingEnabled(): Promise<boolean> {
```
bohandley, line 229: Non blocking, but does setting the flag in conf file evaluate as true? I ran into this issue with tree filters. Customers wanted to set the flag themselves and it would only flip for a value of "treatment" while conf evaluate to true.

bohandley (approved):
LGTM for a/b testing! One comment, non blocking.

PR #1352, opened 2026-07-22
fix(deps): resolve pnpm audit vulnerabilities

shelldandy:
## Summary
- Updates pnpm overrides in `pnpm-workspace.yaml` to resolve `pnpm audit` findings
- brace-expansion >=5.0.7 (high — DoS via exponential-time expansion)
- dompurify >=3.4.12 (low — CUSTOM_ELEMENT_HANDLING bypass)
- fast-uri >=3.1.4 (high — host confusion vulnerabilities)
- immutable >=5.1.8 (high — trie overflow & hash-collision DoS)
- js-yaml >=4.3.0 (high — merge-key chain quadratic CPU)
- protobufjs >=8.6.6 (moderate — prototype mutation & infinite loop)
- webpack-dev-server >=5.2.6 (moderate — cross-origin exposure, CSRF, DoS)

## Test plan
- [x] `pnpm audit` passes with 0 vulnerabilities

PR #1360, opened 2026-07-28
fix(deps): resolve pnpm audit vulnerabilities

shelldandy:
### ✨ Description

**Related issue(s):** n/a — follow-up to #1352

Bumps two `pnpm-workspace.yaml` overrides to close the **2 high-severity advisories** that remain open on `main` after #1352, and fixes an override-precedence bug that made one of those bumps silently inert.

`main` currently reports **5 advisories (3 moderate, 2 high)**. This PR takes that to **3 (3 moderate)** — the 2 high are fixed here; the 3 remaining moderate are the react-router set that needs a breaking v7 upgrade (see below).

| Package | `main` resolves | This PR | Advisory | Severity |
| --- | --- | --- | --- | --- |
| `brace-expansion` | 5.0.7 | **5.0.8** | [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg) (`<= 5.0.7`) | high |
| `postcss` | 8.5.13 | **8.5.23** | [GHSA-r28c-9q8g-f849](https://github.com/advisories/GHSA-r28c-9q8g-f849) (`<= 8.5.17`) | high |

> **Note:** this branch predates #1352 and originally carried the full override sweep. #1352 landed the bulk of it independently, so after merging `main` the unique content here is just the three override lines below. The earlier description of this PR described the pre-merge scope and was stale; this replaces it.

### 📖 Summary of the changes

The whole diff is 3 lines of `pnpm-workspace.yaml` (plus the regenerated lockfile):

```diff
-  brace-expansion@^5: ^5.0.7
+  brace-expansion@^5: ^5.0.8
-  postcss: ^8.5.10
+  postcss: ^8.5.18
-  brace-expansion: '>=5.0.7'
+  brace-expansion: ^5.0.8
```

**1. `postcss: ^8.5.10` → `^8.5.18`.** The old floor looks recent but resolved to **8.5.13**, which is inside the advisory's `<= 8.5.17` vulnerable range. Raising the floor moves it to 8.5.23.

**2. `brace-expansion` — an inert override.** `main` has *both* a version-scoped `brace-expansion@^5` key and a bare `brace-expansion` key. The **bare key wins**, so bumping only `@^5` to `^5.0.8` would have had no effect — the tree stayed pinned at the vulnerable 5.0.7. Both keys are now at `^5.0.8`, and the resolved version is verified at 5.0.8 everywhere. This is the part most likely to be missed in review: the version-scoped bump alone is not sufficient.

**3. `>=` → `^` on the bare key.** Tightened to cap the major, matching the caret style used throughout the file.

**Deliberately not changed:** `webpack-dev-server` stays at `main`'s `'>=5.2.6'` (resolves to 6.0.0). This branch originally capped it at `^5.2.6`; that reverts a working major and pulls the express 4.x subtree back in (~630 lines of extra lock churn) with no security difference, since both satisfy all four wds advisories. Worth noting `@rspack/dev-server@1.1.5` pins wds `5.2.2` exactly, so either override already forces it off its pin — and `pnpm dev` runs `rspack build --watch`, not the dev server.

**Lockfile** was regenerated from `main`'s lockfile rather than this branch's, keeping the diff to 78 lines that are all `brace-expansion` 5.0.7→5.0.8 and `postcss` 8.5.13→8.5.23 plus re-keyed dependents (`nanoid`, `icss-utils`, `postcss-modules-*`). No unrelated dependency float.

### 🧪 How to test?

Verified locally on the merge commit:

- [x] `pnpm install --frozen-lockfile` — passes (the check CI runs; proves the lock matches the merged overrides)
- [x] `pnpm audit` — **5 advisories on `main` → 3 here**; the 2 high are gone. Confirmed by auditing `main`'s manifest in isolation for the baseline.
- [x] Resolved versions verified: `brace-expansion` 5.0.8 (all copies), `postcss` 8.5.23, `webpack-dev-server` 6.0.0, `express` 5.2.1 (unchanged from `main`)
- [x] No duplicate keys in the `overrides:` block — both sides of the merge had added `webpack-dev-server` at different list positions, which git merged into a duplicate YAML key; deduplicated
- [x] `pnpm run typecheck` (`tsc --noEmit`) — passes
- [x] `pnpm run build` (rspack production) — passes, only the pre-existing asset-size warning
- [ ] CI green

**Remaining 3 moderate advisories (out of scope, unchanged from `main`):** [GHSA-wrjc-x8rr-h8h6](https://github.com/advisories/GHSA-wrjc-x8rr-h8h6) and [GHSA-337j-9hxr-rhxg](https://github.com/advisories/GHSA-337j-9hxr-rhxg) need `react-router >= 7.18.0`; [GHSA-jjmj-jmhj-qwj2](https://github.com/advisories/GHSA-jjmj-jmhj-qwj2) names `react-router-dom >= 6.30.5`, which **was never published** (the `version-6` dist-tag still points at the flagged 6.30.4), so no override can satisfy it. The 6.x tree is required by `@grafana/ui` and `@grafana/scenes`, so moving to 7.x is a breaking cross-package upgrade that should be tracked separately. No audit ignores or suppressions were added.


bohandley (approved):
❤️

PR #1364, opened 2026-07-31
test(e2e): force queryLibrary off so saved-queries specs run on nightly

shelldandy:
### ✨ Description

**Related issue(s):** Fixes the red `e2e grafana-enterprise@nightly` job, e.g. [run 30644258327](https://github.com/grafana/metrics-drilldown/actions/runs/30644258327/job/91212682292).

Following: https://github.com/grafana/logs-drilldown/pull/2019/changes#diff-7fbfca66a3aa251082a935adc90ac393956e16f2ccc23c72642e83fadaca2a97

PR #1366, opened 2026-08-04
chore(faro): duplicate rudderstack events as faro events and enhance instrumentation

eskirk:
### ✨ Description

**Related issue(s):** 

we (frontend observability) are improving our faro dogfooding visibility across grafana plugins: this duplicates the plugin's existing rudderstack (`reportInteraction`) events into its faro instance and adds a few more custom events, so user interactions can be analyzed alongside the errors and web vitals faro already collects. it also checks for config that should improve quality of life when using faro/frontend o11y.

### 📖 Summary of the changes

- new `interactionEchoBackend.ts`: registers an echo backend (public `registerEchoBackend()` API) that forwards this plugin's `grafana_explore_metrics_*` interaction events to faro via `pushEvent`, with stringified attributes. registered right after faro init, so it's a no-op when faro is disabled
- new events (kept in a separate commit so it's easy to drop): `datasource_changed`, `app_error_boundary_shown` (crash vs chunk_load), and `selected_metric_action_clicked` with a new `panel_menu_explore` action for the untracked panel-menu explore path
- faro config: `isolate: true` and the page-url `beforeSend` filter were already present — no changes needed. `experimental.trackNavigation` was skipped because the installed faro-web-sdk (2.2.4) predates it; worth revisiting after an sdk bump

### 🧪 How to test?

1. run the plugin against an environment where faro initializes (see `getFaroEnvironment`)
2. click around (select a metric, change datasource, open explore from a panel menu)
3. in the network tab, each `grafana_explore_metrics_*` interaction should appear as a faro event in the collector requests, alongside the existing rudderstack call

existing tests pass (`src/shared/logger/faro`: 20/20, touched suites: 71/71); tsc + eslint + prettier clean

### user journey dogfooding

mirroring these interactions into faro also unlocks [critical user journeys](https://ops.grafana-ops.net/a/grafana-kowalski-app/apps/49/critical-journeys) for this app — journeys are built from faro events, so the duplicated interaction events can be used directly as journey steps to measure completion and drop-off of real product flows.


bohandley:
This should pass tests once this is merged! https://github.com/grafana/metrics-drilldown/pull/1364

shelldandy (approved):

PR #1380, opened 2026-08-13
fix: native histogram no data bug

### LINKED ISSUE #1283: Panel queries: improve adaptive metrics UX
*** Edited 8/13/2026

### What is this?

Native histograms with adaptive metrics can break the histogram check in metrics drilldown resulting in the NH being typed as a gauge and not receiving the rate() function. This results in no data. Customers can then query the metric in Explore and see data.

### What can we do?

We can identify a better heuristic for identifying native histograms that handles this case. I suggest first trying to test/probe for missing metadata, no data result and check to see if query expr `sum(rate(native_histogram[rate]))` returns a heat map result. If these conditions are true, we can flip the panel to a native histogram query (rate + heatmap).

bohandley:
### ✨ Description

**Related issue(s):** 
Resolves https://github.com/grafana/metrics-drilldown/issues/1283
| no data | with correct query and heatmap |
|----|----|
|Screenshot 2026-08-13 at 1 30 23 PM|Screenshot 2026-08-13 at 1 30 33 PM|

Native histogram checks are notoriously flaky due to missing metadata and naming convention. We workaround this by checking a post data result, but this is now flaky as well. It misses a case where adaptive metrics have aggregated labels away resulting in no data for the native histogram and the returned result is no longer in the heat map format.  

Now, if we get no data, we do a probe with `sum(rate(native_histogram[rate]))` which WILL return a heatmap result if the metric is a native histogram.

### 📖 Summary of the changes

1. Check no data gauge metrics with a rate query
2. If the results are heatmap format, switch to rate query with heatmap panel
3. Also respect the function config.

- [ ] This pull request was substantially generated with AI assistance ([GenAI policy](https://github.com/grafana/metrics-drilldown/blob/main/docs/genai.md))

### 🧪 How to test?

1. In ops, the metric `adaptive_traces_sigil_forwarder_receive_delay_seconds` in dev-cortex DS is a native histogram that has aggregated buckets. 
2. First, navigate to [here](https://ops.grafana-ops.net/a/grafana-metricsdrilldown-app/drilldown?from=now-1h&to=now&timezone=utc&var-metrics_filters=&var-filters=&var-labelsWingman=%28none%29&layout=grid&filters-rule=&filters-prefix=&filters-suffix=&search_txt=&filter-firing-alerts=&var-metrics-reducer-sort-by=default&filters-recent=&var-ds=2z9d6ElGk&var-other_metric_filters=), without this fix, this metric shows no data
3. Use graft with this branch on ops. The data shows with the metric, as well as a heatmap panel. 

All tests pass.


shelldandy (approved):
lgtm!

PR #1396, opened 2026-08-20
fix: use derived metric type in config presets to fix native histogram bug

### LINKED ISSUE #1228: Configure Prometheus function dialog shows wrong presets for native histograms

bohandley:
### ✨ Description

**Related issue(s):** 
Resolves https://github.com/grafana/metrics-drilldown/issues/1228

BROKEN
Screenshot 2026-08-19 at 7 47 11 PM

FIXED
Screenshot 2026-08-19 at 8 29 57 PM

This is a refactor of the native histogram check. Back in April we [based our check on returned data frames](https://github.com/grafana/metrics-drilldown/issues/1199) but found this failed on some cases where there was no data returned (no activity and no rate function would not return heatmap cells for native histograms). We had to create a new strategy. 

The [new strategy is a pre-query or probe query with rate](https://github.com/grafana/metrics-drilldown/pull/1380), which is more robust. 

Based on this new strategy, this PR uses that probe query metric type to pass into the config presets to correctly show heatmap and quantiles for native histograms.

### 📖 Summary of the changes

- Refactor native histogram detection to use only the probe query strategy.
- Remove the post query data response check for heatmap cells
- Use the metric type in the config presets

Important to acknowledge: The [original issue](https://github.com/grafana/metrics-drilldown/issues/1228) noted a race condition where the presets might show wrongly, the type is changed and a user may be impacted by a race condition. This is a very narrow window now with the new strategy and I have not been able to trigger it. We are leaving this race condition unhandled in this PR because the tradeoff for creating the plumbing is not worth it.

### 🧪 How to test?

- graft this on to ops
- select a native histogram from the metric list [(link to a native histogram on ops here)](https://ops.grafana-ops.net/a/grafana-metricsdrilldown-app/drilldown?from=now-1h&to=now&timezone=utc&var-metrics_filters=&var-filters=&var-labelsWingman=%28none%29&layout=grid&filters-rule=&filters-prefix=&filters-suffix=&search_txt=%20adaptive_logs_gateway_archival_blocking_duration_seconds&filter-firing-alerts=&var-metrics-reducer-sort-by=default&filters-recent=&var-ds=edprtf91hz01se&var-other_metric_filters=).
- select the metric
- open the config function in the metric view metric panel
- see the config presets are heatmap and quantile

### DIFF
--- a/src/shared/GmdVizPanel/GmdVizPanel.tsx
+++ b/src/shared/GmdVizPanel/GmdVizPanel.tsx
@@ -177,7 +177,7 @@ export class GmdVizPanel extends SceneObjectBase<GmdVizPanelState> {
     this.checkMetricMetadata(discardPanelTypeUpdates);
   }
 
-  // Metadata can resolve a type the sync heuristic can't guess at all (histogram, summary), not just swap
+  // Metadata can resolve a type the sync heuristic can't guess (currently only native histograms), not just swap
   // between counter and gauge. Only applied from a generic sync guess: a metric already classified more
   // specifically by name (e.g. classic-histogram) shouldn't be reprocessed here.
   private applyMetadataResolvedType(metricType: MetricType, metricTypeFromMetadata: MetricType, discardPanelTypeUpdates: boolean) {
--- a/src/shared/GmdVizPanel/GmdVizPanel.tsx
+++ b/src/shared/GmdVizPanel/GmdVizPanel.tsx
@@ -171,10 +171,30 @@ export class GmdVizPanel extends SceneObjectBase<GmdVizPanelState> {
   private async onActivate(discardPanelTypeUpdates: boolean) {
     this.buildVizPanel();
 
-    this.subscribeToStateChanges(discardPanelTypeUpdates);
+    this.subscribeToStateChanges();
     this.subscribeToEvents();
 
     this.checkMetricMetadata(discardPanelTypeUpdates);
+  }
+
+  // Metadata can resolve a type the sync heuristic can't guess at all (histogram, summary), not just swap
+  // between counter and gauge. Only applied from a generic sync guess: a metric already classified more
+  // specifically by name (e.g. classic-histogram) shouldn't be reprocessed here.
+  private applyMetadataResolvedType(metricType: MetricType, metricTypeFromMetadata: MetricType, discardPanelTypeUpdates: boolean) {
+    if (metricType !== 'gauge' && metricType !== 'counter') {
+      return;
+    }
+    if (metricTypeFromMetadata === 'gauge' || metricTypeFromMetadata === 'counter') {
+      return;
+    }
+
+    if (metricTypeFromMetadata === 'native-histogram') {
+      // Also flips panelConfig.type to heatmap (unless the panel type is pinned): a metadata-resolved
+      // native histogram needs the same panel switch the probe already applies.
+      this.switchToNativeHistogramHeatmap(discardPanelTypeUpdates);
+    } else {
+      this.setState({ metricType: metricTypeFromMetadata });
+    }
   }
 
   private async checkMetricMetadata(discardPanelTypeUpdates: boolean) {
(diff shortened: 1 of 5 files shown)


shelldandy (approved):
lgtm!

## Suggestions from `prr`

- nitpick (non-blocking): src/shared/GmdVizPanel/GmdVizPanel.tsx:195: The `else` branch here is unreachable in practice: `getMetricType` can only return `'gauge'`, `'counter'`, or `'native-histogram'` when the sync heuristic first classifies the metric as `'gauge'` or `'counter'` (those are the only branches that reach the metadata fetch). The two guards above already handle all three cases, so this path can never fire. It's harmless as a defensive fallback, but worth a comment noting it should be dead — otherwise a future reader may wonder what scenario it covers.
- nitpick (non-blocking): src/shared/GmdVizPanel/__tests__/GmdVizPanel.test.ts:87: The `discardPanelTypeUpdates=true` path through the metadata-resolved native-histogram branch (`applyMetadataResolvedType` → `switchToNativeHistogramHeatmap(true)`) has no dedicated test. The probe path covers `discardPanelTypeUpdates=true` at line 159, so the underlying `switchToNativeHistogramHeatmap(true)` logic is exercised, but there's no test confirming the metadata path also respects the pin. Consider adding a case alongside the existing 87–95 test that mocks `getMetricType` returning `'native-histogram'` and calls `checkMetricMetadata(true)`, asserting `panelConfig.type` stays unchanged.

PR #1397, opened 2026-08-21
fix(VisualRefresh): Fixes for visual refresh

torkelo:
Progress on https://github.com/grafana/metrics-drilldown/issues/1359 

* Fixes incorrect background when feature toggle `grafana.visualDesignRefresh` is enabled 
* Simplifies background logic (now in 1 place). 
* Align background to use canvas in both light & dark theme (light theme canvas has changed since first versiom of Drilldown and is now ok here), also makes it more concistent with logs drilldown 
* When `grafana.visualDesignRefresh` there is background set as the Page provides that the app. 

Before (see corners, and wrong background)

Screenshot 2026-08-21 at 14 05 09

After:

Screenshot 2026-08-21 at 14 03 43

Before (light theme)
Screenshot 2026-08-21 at 14 04 57

After:

Screenshot 2026-08-21 at 14 04 33


torkelo:
@bohandley can you review this ?

bohandley (commented):
Looks like we have some issues with embedded components:

Toggle off
Screenshot 2026-08-26 at 10 24 25 AM

Toggle on
Screenshot 2026-08-26 at 10 22 40 AM

I will need to identify the areas we need to update for embedded to get the styles the right color and will make a PR to your branch for that if that's ok!

bohandley:
These changes here should fix the embedded app issue. https://github.com/torkelo/metrics-drilldown/pull/1

I will also give you permission to write to this app! Thank you for your patience in creating the fork 🙏🏻

torkelo:
@bohandley merged your PR with a fix to always return a background; it is needed for the sticky elements.  

Also changed for embedded it uses the `background.page` token (it's the same as primary when the feature toggle is disabled). 

When embedded it looks like the whole app is wrapped in a border, can we remove that?

bohandley:
> When embedded it looks like the whole app is wrapped in a border, can we remove that?

Yes, let me create that issue and do it as a follow up!
https://github.com/grafana/metrics-drilldown/issues/1410

bohandley (approved):
🚀 Thank you for this work @torkelo! Excited about the visual design refresh 🥳

PR #1401, opened 2026-08-21
fix: histogram breakdown values bug in applying function

### LINKED ISSUE #1190: Histogram breakdown should keep min/max bucket values of parent
When viewing a breakdown of a native histogram, it seems that each label histogram determines it's own min/max values. This makes it impossible to compare across labels.

For example, the following is a histogram for filesystem I/O request size, grouped by cgroup. It's impossible to answer questions such as "which cgroup is causing all these big requests" from the breakdown.

This is with an otel explicit bucket histogram, ingested via opentelemetry -> alloy -> mimir with mimir's
```yaml
limits:
  otel_convert_histograms_to_nhcb: true
```
, meaning that all histograms share the same bucket configuration

Image

bohandley:
### ✨ Description

**Related issue(s):** 
Resolves https://github.com/grafana/metrics-drilldown/issues/1190

Another bug surfaced in metrics drilldown for histograms. This fixes applying the new histogram breakdown function to label values breakdown, which had been defaulting to show a heatmap.

BEFORE - label value panel is heatmap and not connected to the Histogram function for breakdown
Screenshot 2026-08-21 at 1 14 18 PM

AFTER - Applies the selected histogram breakdown function and shows timeseries panel
Screenshot 2026-08-21 at 1 16 16 PM

### 📖 Summary of the changes

- use the selected function from breakdown on label values 
- do not show heatmap panel
- take the query from a label and decompose it to get data for label values

### 🧪 How to test?

1. graft this branch onto ops.
2. select a histogram metric
3. breakdown by label such as "cluster"
4. see that the label values breakdown uses the selected histogram function and does not show a heatmap panel.

### DIFF
--- a/src/MetricScene/Breakdown/HistogramBreakdownFnVariable.tsx
+++ b/src/MetricScene/Breakdown/HistogramBreakdownFnVariable.tsx
@@ -19,7 +19,7 @@ export class HistogramBreakdownFnVariable extends CustomVariable {
   constructor() {
     super({
       name: VAR_HISTOGRAM_BREAKDOWN_FN,
-      label: t('breakdown.histogram-fn.label', 'Histogram breakdown function'),
+      label: t('breakdown.histogram-fn.label', 'Histogram function'),
       query: HISTOGRAM_BREAKDOWN_FN_OPTIONS.map(({ label, value }) => `${label} : ${value}`).join(','),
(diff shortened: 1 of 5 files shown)


shelldandy (approved):
lgtm!

PR #1421, opened 2026-09-02
fix(VisualRefresh): Fixes background issue in visual refresh

torkelo:
Fixes an issue with the border radius in visual refresh, the top level DataTrail should not have any background (when not embedded) as it interferes with the page border 
Screenshot 2026-09-02 at 11 12 13


bohandley (approved):
🚀

PR #1427, opened 2026-09-10
fix(deps): align react-router with Grafana runtime

shelldandy:
## Summary
- Aligns `react-router` and `react-router-dom` with Grafana's externalized v6 runtime
- Uses published v6.30.6, which backports the GHSA-jjmj-jmhj-qwj2 fix
- Normalizes ErrorView's URL-derived pathname to exactly one leading slash before navigation

## Security rationale
A v7/v8 compile target can mismatch the v6 router supplied by Grafana. GHSA-wrjc-x8rr-h8h6 has no v6 package fix, so the only production `navigate()` call receiving URL-derived input is hardened to remain internal. GHSA-337j-9hxr-rhxg is SSR hydration-only; this browser plugin does not use SSR.

## Test plan
- [x] `pnpm install --no-frozen-lockfile`
- [x] New routing tests: 5/5 pass and Jest exits 0
- [x] Related error-boundary assertions: 24/24 pass; Jest then hit the existing React `MESSAGEPORT` shutdown leak and was terminated after the 600s timeout
- [x] `pnpm run typecheck`
- [x] `pnpm run build`
- [x] ESLint and Prettier checks
- [x] `pnpm audit` leaves only GHSA-wrjc and SSR-only GHSA-337j

## Known remaining
- GHSA-wrjc-x8rr-h8h6: no compatible v6 package patch; vulnerable call site hardened
- GHSA-337j-9hxr-rhxg: not applicable to this non-SSR plugin

This PR remains draft pending review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,8 +59,8 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-inlinesvg": "^4.2.0",
-    "react-router": "^8.3.0",
-    "react-router-dom": "^7.18.1",
+    "react-router": "^6.30.6",
+    "react-router-dom": "^6.30.6",
     "rxjs": "7.8.2",
     "tslib": "^2.8.1",
     "tsqtsq": "^1.1.0"
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-inlinesvg": "^4.2.0",
-    "react-router": "^6.30.6",
     "react-router-dom": "^6.30.6",
     "rxjs": "7.8.2",
     "tslib": "^2.8.1",
--- a/src/App/routingUtils.ts
+++ b/src/App/routingUtils.ts
@@ -1,4 +1,5 @@
-// React Router v6 can interpret protocol-relative and backslash-prefixed values as external URLs.
+// React Router 6.30.6 fixes GHSA-jjmj-jmhj-qwj2 but remains affected by
+// GHSA-wrjc-x8rr-h8h6, which can treat backslash-prefixed values as external URLs.
 export function normalizeInternalPathname(pathname: string): string {
   return `/${pathname.replaceAll('\\', '/').replace(/^\/+/, '')}`;
 }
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -59,8 +59,8 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-inlinesvg": "^4.2.0",
-    "react-router": "^8.3.0",
-    "react-router-dom": "^7.18.1",
+    "react-router": "^6.30.6",
```
L2D2Grafana: Interesting I'm pretty sure you only need react-router-dom. 
react-router-dom re-exports (useLocation, useNavigate, Routes, Route, Navigate, MemoryRouter), so metrics could switch all 6 import sites to 'react-router-dom' and drop the direct react-router dependency entirely

src/App/routingUtils.ts
```diff
@@ -0,0 +1,4 @@
+// React Router v6 can interpret protocol-relative and backslash-prefixed values as external URLs.
+export function normalizeInternalPathname(pathname: string): string {
+  return `/${pathname.replaceAll('\\', '/').replace(/^\/+/, '')}`;
+}
```
L2D2Grafana, line 5: This is nice, but I'm not sure it's needed. The updated v6 should have the security update.

L2D2Grafana (dismissed):

package.json
shelldandy: Updated in 5ad408dd: all six app and test import sites now use `react-router-dom`, and the redundant direct `react-router` dependency and lockfile importer entry are removed.

src/App/routingUtils.ts
shelldandy: React Router 6.30.6 patches GHSA-jjmj-jmhj-qwj2, but GitHub still lists v6 as affected by the separate GHSA-wrjc-x8rr-h8h6 backslash/open-redirect bypass (first patched in 7.18.0). I kept this defense-in-depth normalization and clarified the advisory distinction in 5ad408dd.

package.json
```diff
@@ -59,8 +59,7 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-inlinesvg": "^4.2.0",
-    "react-router": "^8.3.0",
```
L2D2Grafana, line 62: nice!

L2D2Grafana (approved):
Way to remove react-router 😎

PR #1446, opened 2026-09-21
fix(theme): use accent token for secondary actions

### LINKED ISSUE #1442: chore: migrate primary-color usages to the accent token
### what is this?
The issue reserves theme.colors.primary for one primary action per page, replacing its use in checkbox, switch, and focus styles with a new orange accent token.

### what to do
- WithConfigPanelOptions.tsx (focus outline, selected and hover border) and LoadQueryModal.tsx (focus-visible outline) match the issue's named examples directly.
- AttributeDistribution.tsx's valueRowIncluded hover and selected highlight (colorManipulator.alpha of theme.colors.primary.main at 0.15 and 0.25) is the same selected/hover pattern as the two files above.
- The chart-bar fill color in the same file (theme.colors.primary.main as a fallback, line 827) is not in scope. The issue does not name chart or data-viz colors as something to check.

efegokdemir:
Fixes: https://github.com/grafana/metrics-drilldown/issues/1442

## Summary

Use the visual-refresh accent token for secondary actions instead of the primary token, so the controls retain the intended hierarchy in refreshed themes.

## Changes

- Updated the affected secondary action styles to use the accent token.
- Reconstructed the existing contribution on current `main` as a DCO-signed, cryptographically signed commit.

## Testing

- Existing PR checks were green before the signing-only history repair.
- `git diff` and `git range-diff` confirm the implementation tree and patch are unchanged.
- GitHub reports the replacement commit signature as verified.

## Notes

AI assistance was used for implementation and validation; I reviewed the submitted change and take responsibility for it. The maintainer merge-from-main commit carried no unique tree changes and was safely replaced by the linear current-main reconstruction.


efegokdemir:
Repaired the existing branch for the signing requirement. The implementation tree and patch are unchanged per git diff/range-diff; the contribution is now DCO signed and GitHub-verified. Existing checks were green before this signing-only history repair and will rerun on the updated branch.

bohandley (approved):
LGTM! Changing from primary to accent for secondary actions

Image
Image

PR #1451, opened 2026-09-21
feat(related): show related metrics count

### LINKED ISSUE #1330: UX: add count to related metrics tab
#### What is this?

The related metrics tab does not have a count. The related logs tab does. This is a UX inconsistency.

Image

#### What to do

Show a count for related metrics.

efegokdemir:
### ✨ Description

**Related issue(s):** https://github.com/grafana/metrics-drilldown/issues/1330
Resolves https://github.com/grafana/metrics-drilldown/issues/1330

The Related metrics tab now displays the number of metrics returned by the current filtered result set before the tab is selected, matching the existing Related logs tab lifecycle. The label intentionally describes the value as returned results rather than an exact backend total: the variable API exposes options but no total or truncation metadata, and the repository's 10,000 frontend hardening limit applies to ad hoc filter providers, not this metric query.

### 📖 Summary of the changes

- Preload and reuse the related-metrics scene after the active tab finishes loading, so the count is available before selection without a second request.
- Track the filtered related-metrics option count on `MetricScene`.
- Update the count when related-metrics options load or change through search and filters.
- Clarify the tab tooltip so the displayed value is not presented as an exact total.
- Add focused coverage for loading, returned counts, filtering, and the 10,000-result boundary.

### 🧪 How to test?

- `corepack pnpm test:ci` — 80 suites, 890 tests passed; one worker was force-exited after tests completed.
- `corepack pnpm exec jest src/MetricScene/RelatedMetrics/RelatedMetricsScene.test.ts src/App/Trail.test.tsx src/MetricsReducer/metrics-variables/__tests__/MetricsVariableFilterEngine.test.ts --runInBand --forceExit` — 3 suites, 36 tests passed.
- `corepack pnpm typecheck` — passed.
- `corepack pnpm lint` — passed with 19 pre-existing warnings and no errors.
- `corepack pnpm build` — passed with the repository's existing asset-size warning.
- `corepack pnpm exec prettier --check src/MetricScene/MetricActionBar.tsx src/MetricScene/MetricScene.tsx src/MetricScene/RelatedMetrics/RelatedMetricsScene.tsx src/MetricScene/RelatedMetrics/RelatedMetricsScene.test.ts` — passed.
- `git diff --check` — passed.

Both rewritten contribution commits are DCO signed and cryptographically signed with the registered SSH signing key; GitHub reports both signatures as verified.


joey-grafana (changes requested):
We only show the related metrics count if click the tab whereas we show the count on the logs tab before it is clicked. We should be consistent here. Also, it's at 10,000 results, not sure if that's a hard limit or just what the API returns. cc @bohandley in case I'm missing something here 

Image

efegokdemir:
Updated the existing branch. Related metrics are now preloaded after the active tab completes, so the count is available before selecting the tab and the same scene is reused when selected, avoiding a second request. The count follows the loaded/filtered options. I also verified that the repository’s 10,000 frontend hardening limit is for ad hoc filter providers; this metric variable exposes no total or truncation metadata, so the tooltip now describes the value as the number returned rather than an exact total. Focused tests, typecheck, lint, and formatting pass; the new commit is DCO signed and GitHub-verified.

efegokdemir:
Follow-up to the signing correction: the original unsigned implementation commit was also rewritten, so both contribution commits on the existing branch are now DCO signed and GitHub-verified. The final tree and patch remain identical per git diff/range-diff. The implementation and validation details in the PR description are unchanged.

PR #1452, opened 2026-09-21
fix(search): make metric name search case-insensitive

efegokdemir:
## Summary

Make metric-name search case-insensitive while preserving the existing filtering behavior.

## Changes

- Updated the metric name filter to match names without case sensitivity.
- Preserved the existing test coverage and reconstructed the contribution on the current `main` base as a DCO-signed, cryptographically signed commit.

## Testing

- Existing PR checks were green before the signing-only history repair.
- The rewritten commit has the same tree and implementation patch as the pre-rewrite branch, verified with `git diff` and `git range-diff`.
- GitHub reports the new commit signature as verified.

## Notes

AI assistance was used for implementation and validation; I reviewed the submitted change and take responsibility for it. The maintainer merge-from-main commit was not carrying unique tree changes and was safely replaced by the linear current-main reconstruction.


shelldandy (approved):
lgtm!

shelldandy:
@efegokdemir make sure all your commits are signed please

shelldandy (changes requested):
ensure all commits are signed

efegokdemir:
Repaired the existing branch for the signing request. I preserved the exact implementation tree and patch with git diff/range-diff, reconstructed the contribution on current main, and pushed with force-with-lease. The replacement commit is DCO signed and GitHub reports its signature as verified. Existing checks were green before the signing-only rewrite; the branch checks will rerun.

joey-grafana (approved):

